PR #29, opened 2025-01-11
State machine

avidraccoon:


avidraccoon:
@jkleiber the branch this is based off of does not have spotless what do I do?

sleepyghost-zzz:
@avidraccoon  i think this is what u do, give it a try:
paste: id "com.diffplug.spotless" version "6.12.0" into plugins
then throw this somewhere in the gradle file:
spotless {
	// optional: limit format enforcement to just the files changed by this feature branch
	ratchetFrom 'origin/main'

	format 'misc', {
		// define the files to apply `misc` to
		target '*.gradle', '.gitattributes', '.gitignore'

		// define the steps to apply to those files
		trimTrailingWhitespace()
		indentWithTabs() // or spaces. Takes an integer argument if you don't like 4
		endWithNewline()
	}


hope this helps

src/main/java/frc/robot/subsystems/StateMachineExample.java
```diff
@@ -0,0 +1,101 @@
...
+
+import coppercore.controls.state_machine.StateMachine;
+import coppercore.controls.state_machine.StateMachineConfiguration;
+import coppercore.controls.state_machine.state.PeriodicStateInterface;
+import coppercore.controls.state_machine.state.StateContainer;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+
+public class StateMachineExample extends SubsystemBase {
+
+    static class IdleState implements PeriodicStateInterface {
```
aidnem, line 11: @jkleiber @avidraccoon I seem to recall us wanting to put each state in its own file. Do y'all think this is the direction we should go this year?
In my opinion we should because:
- Smaller files, easier to find stuff within files
- Removes the need for 3 levels of nesting within this file (methods implemented within a class within a class)
- Less overwhelming
- Easier to jump to files than functions in my opinion
- Separation of responsibility
- We can put each state in a folder under each subsystem to keep them organized anyway

avidraccoon: Yes, do you want me to just make a not3 in the code the states should be in separate files or move them to be in separate files?

jkleiber: Let's just move them to be in separate files to have a good example for future state machines

build.gradle
```diff
@@ -1,42 +1,68 @@
 plugins {
-    id "java"
-    id "edu.wpi.first.GradleRIO" version "2025.1.1-beta-1"
+	id "java"
+	id "edu.wpi.first.GradleRIO" version "2025.1.1-beta-1"
```
jkleiber, line 3: This should be 2025 non-beta

@aidnem or @linglejack could you take a pass at build gradle to make sure other things aren't missing?

build.gradle
```diff
@@ -1,42 +1,68 @@
...
+	id "com.diffplug.spotless" version "6.12.0"
```
aidnem, line 4: The build.gradle on my elevator branch is using version 6.24.0 instead of 6.12.0, I assume the newer version is what we want to go with.

build.gradle
```diff
@@ -46,35 +72,56 @@ wpi.java.debugJni = false
 
 // Set this to true to enable desktop support.
 def includeDesktopSupport = true
+repositories {
+	mavenLocal()
+	mavenCentral()
+	maven {
+		name "photonvisionRepositoryReleases"
+		url "https://maven.photonvision.org/releases"
```
aidnem, line 80: I don't think this depends on photonvision so I don't think this needs to be here, we might as well let it get added with the vision implementation.

build.gradle
```diff
@@ -46,35 +72,56 @@ wpi.java.debugJni = false
...
-    nativeRelease wpi.java.vendor.jniRelease(wpi.platforms.desktop)
-    simulationRelease wpi.sim.enableRelease()
+	nativeRelease wpi.java.deps.wpilibJniRelease(wpi.platforms.desktop)
+	nativeRelease wpi.java.vendor.jniRelease(wpi.platforms.desktop)
+	simulationRelease wpi.sim.enableRelease()
+
+	testImplementation 'org.junit.jupiter:junit-jupiter:5.10.1'
+	testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
+
+	implementation 'com.github.team401:coppercore:57a9a59be7'
```
aidnem, line 115: Could you use maven local to test this instead of Jitpack?

build.gradle
```diff
@@ -46,35 +72,56 @@ wpi.java.debugJni = false
...
+}
 
-    testImplementation 'org.junit.jupiter:junit-jupiter:5.10.1'
-    testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
+repositories {
+	maven { url 'https://jitpack.io' }
```
aidnem, line 119: We should also remove this jitpack repository addition (see comment on line 115)

build.gradle
aidnem: Something like this would replace it:
```groovy
	def coppercoreVersion = "2025.0.0-beta"
	implementation "io.github.team401.coppercore:controls:$coppercoreVersion"
```

build.gradle
aidnem: @jkleiber I compared this build.gradle with the one on my branch 4-add-elevator, looks like this one is an older version: using older spotless, still importing from jitpack. I commented everything that stood out to me as being different/incorrect.

build.gradle
avidraccoon: I believe it was throwing an error without that

sleepyghost-zzz:
@avidraccoon idk if you realize but you deleted spotless, so you won't be able to build

jkleiber:
This PR demonstrated state machine from coppercore works. Closing now that verification is done

Issue #31, opened 2025-01-14
Add robot rotation lock to sides of reef


linglejack06:
**Purpose**
The purpose of this addition is to help auto alignment with scoring in teleop and auto progressed by otf path follow

**Project Scope**
- [ ] Add tag numbers with some sort of control to select them
- [ ] override rotation from joystick to lock onto specified tag id

aidnem:
@jkleiber @linglejack06 Do we want to lock onto the direction of the side of the reef instead of pointing directly at the tag? This way we'd stay aligned if we have to be off center to score (I'm not sure how the robot will look but I don't think we can stay 100% centered on the reef side and still be able to score).

Maybe we could do this only when we're within a certain distance to make sure that we stay pointed at the tag while far away.

Benefits are as follows
- Easier to align (just switch to a heading)
- On-the-fly can just be provided with an end goal heading to line up to the side of the reef
- Less rotation while strafing around the reef, resulting in less rotation while driving and therefore less drive skew (not sure if this was an issue last year or not)

jkleiber:
@aidnem yeah the idea of this ticket is to basically have a mode where we are lined up directly facing the relevant side of the reef for scoring purposes after we OTF over to a scoring position

sleepyghost-zzz:
@jkleiber hi, i was curious about 2 things. one, can i get an updated definition of what i should be doing, and 2, can you look at the (WIP) commit and help me figure out why it isnt turning?

linglejack06:
@sleepyghost-zzz fyi there's code for aligning to a specific rotation point in high key 2024 ( somewhere in drive) and in 2024 robot code (again in drive, I think periodic or run velocity equivalent )

linglejack06:
That code might help some, also check chief delphi for lots of examples

jkleiber:
@sleepyghost-zzz can you point me to a PR with your changes. It would make it easier to review

PR #46, opened 2025-01-21
add OTF follow to drive

linglejack06:
- add DriveWithJoysticksReefscape command to handle OTF and joystick override
- add isOTF boolean to tell drive when it needs to run pathfinding
- add FindOTFPoseFromLocation method to gather pose info
- add getDriveToPoseCommand to find path following command and schedule / cancel in periodic


linglejack06:
@jkleiber @aidnem heres a basic setup of command and then enum representing OTF locations. i had somequick questions before i continued: 

1. how do we want to set drive to OTF mode, like what button? this might be a @minhnguyenbhs question
2. also, how do we want to keep track of OTF locations button wise?

Once i figure these out, i think it should be a relatively quick PR, shouldn't have much logic additions, and i should maybe be able to test tonight. will you be here tonight Aiden?

avidraccoon:
Do the joysticks not return a value from -1 to 1?

linglejack06:
> Do the joysticks not return a value from -1 to 1?

yes, good catch regarding command

linglejack06:
@jkleiber do you prefer drive class to find pose from location, or should we have a command that just says for example if button 1 then its this pose. set drive to this pose

drive 
```
switch(this.location) {
case Reef1:
return Pose2d(location coords)
}
```

or command
```
public void execute () {
if (controller.a.isPressed()) {
drive.setOTFPose(new Pose2d(reef1 coords)
}
}
```

linglejack06:
currently all set up in drive, but it would be easy to move. 

***logic goes...***
if OTF true -> find pose for scoring -> make pathfind command based on pose -> schedule pathfind command (run velocity can still override rotation at a later point)

else OTF false -> cancel pathfind command (ensuring that it is not setting any goal speeds), this is also ensured by drive with joysticks sending goal speeds again with `super.execute()`

jkleiber:
@linglejack06 I don't think there are hard requirements on where the OTF path goal is set. 

We need to make a design for how the OTF pose gets selected during teleop. I think it would be a good idea to make a short slide deck that shows how buttons map all the way through to OTF pose selection. Then it might become more clear where the logic should live. I'll leave that design work up to you and will provide feedback.

I'd consider the following possibilities:
- manual selection of the goal by button masher
- auto goal selection by current pose
- driver quick change target (like channel up or down on a TV remote, i.e. Side step the current goal and choose the one on the left or right)
- auto selection without human input (if we don't press any buttons where will it automatically go next?)

I'm not asking you to implement these things in this PR. However, by creating a design for how it might all work together it should become clear what a good way to implement could be

linglejack06:
@jkleiber  i figured out DriveWithJoysticksReefscape will be useless. When going in otf pathplanner takes over as a command, therefore, drive with joysticks wont be executed (so no switching back to manual mode)

therefore, the option i say we pursue is make driver hold trigger down on joystick for otf, when trigger is let go, path follow command is cancelled and control automatically goes back to DriveWithJoysticks.

alternatively, we just move the joystick input logic into robot container periodic, but i really dont like that clutter and feel that driver holding a trigger makes it more clear to the driver that robot is automatically path following (less accidental cancelling)

linglejack06:
@jkleiber @aidnem also, heres my slide deck for proposed Controls, i feel this is very intuitive. It would just require the computer to highlight current scoring, processor and reef level locations (which shouldnt be hard)

[https://docs.google.com/presentation/d/1Caos9YEPjYjOtV4V0XzVERPZ-C7mOEJP_OzqO0l3F98/edit?usp=sharing](url)

avidraccoon:
Just so you know the email was set up wrong in markdown  here is the [fixed link](https://docs.google.com/presentation/d/1Caos9YEPjYjOtV4V0XzVERPZ-C7mOEJP_OzqO0l3F98/edit?usp=sharing)

aidnem:
One question about using a pathplanner path for the final landing zone: Do we want to manually use a single tag for scoring lineup [as explained in this post from MA](https://www.chiefdelphi.com/t/frc-6328-mechanical-advantage-2025-build-thread/477314/85?u=pchild) or are we going to trust global localization to be accurate enough for scoring?

linglejack06:
thats one of the primary things i was asking, i think if we feel our vision
pose estimate globally wont be strong enough, we lineup based on the tag,
which might honestly be simpler either way

jkleiber:
@linglejack06 Nice work on the slide deck, I think the approach makes sense. I have two points of feedback on the controller side:
1. I think we probably want to have a button for the masher to press to select the closest reef scoring location to the robot so they can then go left/right from there rather than having to toggle through the whole reef. Additionally, the up/down arrow on the d-pad could be used to select the reef location 180 degrees across from the current selection. Obviously we'll need to start developing a way to visualize this in elastic - @minhnguyenbhs could you create a ticket and add it to your UI milestone?
2. The driver should be able to do the left/right thing as well using the little d-pad thing on top of the joystick. If they are lined up on an already scored coral location and want to side step, they could switch the target without masher input. I think we would want that to only be available in #31 mode though

@aidnem My expectation is that OTF will get us to the final approach before we engage a precision maneuver to score a coral. So I think the baseline expectation is that we will do some single tag tracking in #31. In the event that global localization is bad, we need to be able to skip OTF paths and just go direct to #31 based on whatever tag is in view. We can do that in #31 though.

linglejack06:
For 1. If we used up down as opposite side, we would then lose access to the reef level, how do you propose we select reef level in that scenario @jkleiber 

For 2. I agree completely, that will probably be a little complex in code, as typically when switching scoring locations we would go back to OTF, but I'm sure a few simple conditionals would fix this issue (just by checking if we already OTF)

Regarding global tracking, I think the idea would be that we trust global tracking enough to get us to the lineup pose. We then lock on to that sides single tag and lineup with it. Because that's such a fine movement (+/- 1.5 inch I think) we can't rely on global tracking for the "landing zone"

linglejack06:
Also, are you fine with the approach regarding the joystick trigger?

jkleiber:
@linglejack06 good point about the reef level - for now we can put the across-reef button on the shelf and think about it later. Maybe a bumper would be the way to go there.

The trigger on the joystick is a good starting place at the very least. I think we should go with that for now and reassess if it turns out that it doesn't work. My only concern is that we probably want to use that trigger to do the full scoring sequence, and there are cases where we will want to skip OTF manually (i.e. if the driver already got us to the lineup position)

linglejack06:
Ok, sounds good.

For trigger, we could probably always do a check before setting otf, like supplier from drive that says is drive pose within certain distance of the desired on the fly location if it is we skip otf and move straight to landing zone lineup. I think that's a problem more for #31 though, once we can get basic otf working

jkleiber:
Agreed @linglejack06 

One point to note - if we are skipping OTF, it might be because global pose is bad. In which case we can't trust the check against pose vs OTF destination. That being said, it will be a good check to have for auto-skipping OTF in cases where we are already close enough

I think we'll probably need another button to force the skip, but yet again we can cross that bridge in #31

linglejack06:
@jkleiber whenever i run the pathfinding command, we have a 80 ms loop overrun and i cant drive robot anymore, neither does it drive the path. Must be an error with processing the path, even though im running warmuo command.

do you have any ideas? or @aidnem have you experienced this?

linglejack06:
loop overrun is 5.6 seconds without running warmup command, i wonder how we cut down on this time, is it reproducible on your computers?

jkleiber:
@linglejack06 the video looks good (when it works) - could we add a rotation target as well so the robot is lined up for intaking from the coral station?

linglejack06:
sure, ill just add that into the otf pose, is it a 60 degree angle?

linglejack06:
poses for coral stations left and right have been added with proper x,y, and rotation. This is working perfectly in sim after pathfinding command gets over loop overrun

jkleiber (approved):

PR #49, opened 2025-01-22
basic functionality- climb in sim

minhnguyenbhs:
Hi! I'm not sure what isn't working here- it's probably a setting on the SingleJointedArmSim, but I'm not sure what it is

{036C6232-E7D1-42A9-AE2D-7E33831D44EC}

The state machine is presumably working since it's going from None to Climbing + Idle to Searching, and the PID controller is also working since the appliedVoltage value is changing, but applying inputVoltage isn't doing anything to the motor angle

could you look over this for me? Thank you for your help! Climb should be nearly done after this


src/main/java/frc/robot/subsystems/climb/ClimbIOSim.java
```diff
@@ -0,0 +1,79 @@
...
+        if (override) {
+            appliedVolts = overrideVoltage;
+        } else {
+            appliedVolts =
+                    Volts.of(
+                            angleController.calculate(climb.getAngleRads(), goalAngle.in(Radians)));
+        }
+
+        outputs.appliedVoltage.mut_replace(appliedVolts);
+        climb.setInputVoltage(appliedVolts.in(Volt));
```
jkleiber, line 60: I think you need to also call `climb.update(0.02)` where 0.02 is the update rate (we probably have a constant lying around somewhere for this).

I recommend taking a look at the example code for this here: https://github.com/wpilibsuite/allwpilib/blob/main/wpilibjExamples/src/main/java/edu/wpi/first/wpilibj/examples/armsimulation/subsystems/Arm.java

jkleiber: Also, I think we should make a separate function for doing the simulation called updateSimulation() and then call it from either applyInputs or applyOutputs

The argument for doing it in applyInputs is that you can then use the sim to update sensor values

src/main/java/frc/robot/subsystems/climb/ClimbIOSim.java
```diff
@@ -0,0 +1,79 @@
...
+                    DCMotor.getKrakenX60(2),
+                    75,
+                    SingleJointedArmSim.estimateMOI(0.5, 0.5),
+                    0.5,
+                    0,
+                    2 * Math.PI,
+                    false,
+                    0,
+                    0,
+                    0);
```
jkleiber, line 27: You might need to add in a non-zero value for the second to last argument, like is done in the example code: https://github.com/wpilibsuite/allwpilib/blob/main/wpilibjExamples/src/main/java/edu/wpi/first/wpilibj/examples/armsimulation/subsystems/Arm.java

src/main/java/frc/robot/subsystems/climb/ClimbSubsystem.java
```diff
@@ -0,0 +1,123 @@
...
+                break;
+            case LIFTING:
+                lifting();
+                break;
+            default:
+                idle();
+                break;
+        }
+
+        io.updateInputs(inputs);
```
jkleiber, line 115: Can we update the inputs prior to running the state specific logic?

src/main/java/frc/robot/subsystems/climb/ClimbSubsystem.java
```diff
@@ -0,0 +1,123 @@
...
+
+    private BooleanSupplier rampClear = () -> true;
+
+    public enum ClimbAction {
+        NONE, // do nothing
+        CLIMB, // start automated climb sequence
+        OVERRIDE
+    }
+
+    private enum ClimbState {
```
jkleiber, line 28: Could we try using the state machine from coppercore?

jkleiber (commented):
hopefully this is helpful!

minhnguyenbhs:
{0952E797-57B2-46EE-B846-5351005F714F}
The arm is moving now- thank you so much for your help!

src/main/java/frc/robot/subsystems/climb/ClimbSubsystem.java
minhnguyenbhs: Is there an existing example of coppercore state machine anywhere?

minhnguyenbhs:
sorry I'm really really struggling to understand json state machine- what is going on?

jkleiber:
@minhnguyenbhs I don't think the state machine is related to JSON. @avidraccoon would be the person to talk to for both though

minhnguyenbhs:
sorry I keep mixing up the terms! I did ask him about it but I think I'll be able to figure it out later- thank you!

minhnguyenbhs:
@jkleiber I added the java state code! It currently looks like this:

image

minhnguyenbhs:
@aidnem hi when you get the chance could you take a look at motionmagic for me thanks!

minhnguyenbhs:
@aidnem hi!! definitely not trying to rush you haha but were you able to take a look at this today maybe? it's definitely fine if you haven't I was just checking since I'm at shop :D

jkleiber (commented):
Here's a round of review feedback

As a note: I want to get this subsystem done by the weekend so we can start simulating the whole robot. We need to have all the pieces working together before we get to the hardware, so thank you in advance for prioritizing this!

src/main/java/frc/robot/subsystems/climb/ClimbSubsystem.java
```diff
@@ -0,0 +1,119 @@
...
+  // TODO: replace when ramp becomes a thing
+  private BooleanSupplier rampClear = () -> true;
+
+  public enum ClimbAction {
+    NONE, // do nothing
+    START_CLIMB, // start automated climb sequence
+    SYSTEM_READY,
+    READY_FOR_CLIMB,
+    OVERRIDE
+  }
```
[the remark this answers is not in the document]
minhnguyenbhs: can the NONE action maybe act as the CANCEL action?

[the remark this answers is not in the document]
jkleiber: If you rename it to CANCEL I think I would be fine with that

src/main/deploy/constants/comp/FeatureFlags.json
```diff
@@ -1,4 +1,7 @@
 {
-  "runScoring": false,
-  "runDrive": true
+  "runElevator": true,
```
jkleiber, line 2: runElevator doesn't exist here anymore, please remove

src/main/deploy/constants/hitl/FeatureFlags.json
```diff
@@ -1,4 +1,7 @@
 {
+  "runElevator": false,
```
jkleiber, line 2: please remove runElevator

jkleiber (commented):
Some unhelpful comments

jkleiber:
@minhnguyenbhs 

Debugging tips:

1. I recommend opening scope and seeing if the actions/states are changing while the arm fails to move. I also recommend logging goal position and motor output

2. Confirm that you're using the correct robot in config.json and that you're pressing the A button. 

3. I feel like something is weird about passing `instance` to the states in Climb subsystem before instance is initialized. Confirm that this is how we do this in other subsystems to rule this out

minhnguyenbhs:
image

@aidnem arm is moving but less well than pid model

minhnguyenbhs:
I think that the instance is used like this in the elevator branch? Should it be done in a different way?

src/main/java/frc/robot/constants/ClimbConstants.java
```diff
@@ -0,0 +1,58 @@
...
+          new JSONSyncConfigBuilder().setPrettyPrinting(true).build());
+
+  public final Angle restingAngle = Degrees.of(20);
+  public final Angle searchingAngle = Degrees.of(60);
+  public final Angle finalHangingAngle = Degrees.of(0);
+
+  public final Integer leadClimbMotorId = 0;
+  public final Integer followerClimbMotorId = 1;
+
+  public final Double climbkS = 160.0;
```
jkleiber, line 27: This is a really really high value for kS. Suggest making this and other feedforwards 0 to help you debug

If they aren't being used, still a good idea to 0 these out

jkleiber (commented):
I'm nearly ready to approve this. I have a comment on feedforwards, and also there is a merge conflict. Once you resolve the merge conflict and make the change to feedforward I will probably be ready to approve

src/main/java/frc/robot/InitBindings.java
```diff
@@ -124,4 +126,15 @@ public static void initDriveBindings(Drive drive) {
                 },
                 drive));
   }
+
+  public static void initClimbBindings(ClimbSubsystem climb) {
+    driverController
+        .a()
+        .onTrue(new InstantCommand(() -> climb.fireTrigger(ClimbAction.CLIMB)))
+        .onFalse(new InstantCommand(() -> climb.fireTrigger(ClimbAction.NONE)));
```
aidnem, line 134: I don't think you need an onFalse here right? Does ClimbAction.NONE do anything?

jkleiber: Yeah it probably makes sense to remove this @minhnguyenbhs because you could get rekt by frame perfect stuff relating to letting go of the button right as the subsystem decides it is time to proceed

src/main/java/frc/robot/constants/ClimbConstants.java
minhnguyenbhs: I switched them to 0! Since it's hard to simulate hanging on a bar, do you think these will be used in the actual robot?

jkleiber (approved):
ship it

good work @minhnguyenbhs

PR #53, opened 2025-01-23
fix drivetrain constants with static field in json constants

linglejack06:
- add drivetrain constants static field to json constants class


linglejack06:
@jkleiber json constants pretty much already implemented, just not used everywhere. i tested sim and robot is running, so should be good, based off what i saw in json file it looked correct (other than being 1200 lines long)

just need the environment thing aiden is working on to be able to tune test bot

linglejack06:
@jkleiber all gains were lost relating to pid though, not sure why they arent making it over, probably issue with saving data?

jkleiber:
@linglejack06 yeah we need to regenerate the JSON file for the drivetrain constants so that the data we have is actually correct rather than default values

linglejack06:
Ok, I’ll try again and see if that works. Sent from my iPhoneOn Jan 23, 2025, at 4:23 PM, Justin Kleiber ***@***.***> wrote:﻿
@linglejack06 yeah we need to regenerate the JSON file for the drivetrain constants so that the data we have is actually correct rather than default values

linglejack06:
@jkleiber its still saving with 0 gains on everything, we might need to seperate the gains into variables and then assign ie
```
 public static final double driveKP = 0.4;
drivegains.setkP(driveKP);
```

linglejack06:
@jkleiber it appears that the gains are the only things not showing up, everything else in the module creator area is populating (drive gear reduction, steer, etc)

jkleiber:
@linglejack06 the gains probably need to be defined separately as you showed in your example because they're set via a function rather than being a variable. So I'd try that and see if it works

linglejack06:
Nathan actually solved last night, we just need to remove static field  from the gains and they work perfectlySent from my iPhoneOn Jan 23, 2025, at 8:39 PM, Justin Kleiber ***@***.***> wrote:﻿
@linglejack06 the gains probably need to be defined separately as you showed in your example because they're set via a function rather than being a variable. So I'd try that and see if it works

jkleiber:
I think this PR is a duplicate of some of the work I've been asking @sleepyghost-zzz to do in #58 . I recommend we go with the stuff in that PR since this one is stale and doesn't seem to be blocking any development. Closing this PR but feel free to re-open if you disagree @linglejack06

PR #56, opened 2025-01-24
Add final landing zone lineup at reefs

### LINKED ISSUE #17: Ensure vision works properly with drivetrain developed in #1
**Purpose**
Make sure vision does not cause robot to estimate itself as much faster than truth

**Project Scope**
- [x] add vision into robot container
- [x] test in sim without drive with joysticks command (rule out command)
- [x] test in sim with command to see if everything works together
- [x] fix whatever arises as the issue, or no fix is issue isn't found in this code

### LINKED ISSUE #31: Add robot rotation lock to sides of reef

linglejack06:
- add several methods relating to lineup: find tag id, find camera id, find goal speeds, etc
- add pid controllers for along track, cross track, and rotation in lineup
- add vision in initSubsystems
- add vision constants

Resolves #31 
Resolves #17


linglejack06:
@jkleiber before i move further, this will involve querying vision and then setting goal speeds therefore do we prefer to...

give drive a vision supplier (of current tags seen + distance from them), then have a lineup boolean, set in instant command. drive would then handle finding distance from tag and minimizing that distance + tag offset.

OR do we prefer creating command with vision and drive subsystems as requirements, which would find the distance and call set goal speeds from there

I feel option 1 more closely aligns to what we have done previously (with otf and previous years) but option 2 might be a less cluttered approach.

thoughts? also @aidnem

jkleiber:
I think option 1? I'll describe my take:

If we are lining up to a particular reef pole, we'd know:
1. The relevant tag
2. The relevant camera (assuming 2 forward facing)


So then we'd have vision compute distance to tag and cross track error. We'd target 0 cross track error and the correct distance to the tag (which we could make 0 if we factor in the camera position and apply some offset).

The drivetrain would basically apply robot relative speeds to regulate crosstrack error and distance while in alignment mode within the subsystem

linglejack06:
@jkleiber one issue i came across for aligning with single tag:

we either rely on 3d pnp pose estimate, which would produce ambiguity. Not sure if this error would be ;arger than +/- 1.5 inches. my guess is the error would be insubstantial at these close distances. I feel this is the best option we have due t the complications of the second option. This option would just be gathering the single tag pose estimate from the specific camera and then using that estimate to drive our robot closer to the tag desired.

OR we make a custom algorithm that takes the undistorted corners and finds direct 3d distances based on x and y angles. This would likely be better and is what 6328 has mentioned. However, i would be completely unsure how to implement this with photon vision. 

this might be worth a read: https://www.chiefdelphi.com/t/frc-6328-mechanical-advantage-2025-build-thread/477314/85

the only built in solution with photon vision is finding distance away from tag, not cross track distance. One other solution i found is we could use this, minimizing that distance, and aligning to tag. we then strafe the robot until our rotation aligns with that hexagons side angle (ie when robot is at 60 degrees we know we are perfectly aligned with tag, likely meaning we have little cross track error)

linglejack06:
a third option could be using the 3d photon estimate as described in 1. but only using its translation values to help reduce ambiguity errors. we would then just get robot rotation from pigeon 2

linglejack06:
or maybe we use the transform 3d for camera to target to minimize those values, do you have a preferred method of doing this, i assume all will work but which one would produce least error

linglejack06:
```
        var target = result.targets.get(0);

        // Calculate robot pose
        var tagPose = aprilTagLayout.getTagPose(target.fiducialId);
        if (tagPose.isPresent()) {
          Transform3d fieldToTarget =
              new Transform3d(tagPose.get().getTranslation(), tagPose.get().getRotation());
          Transform3d cameraToTarget = target.bestCameraToTarget;
          Transform3d fieldToCamera = fieldToTarget.plus(cameraToTarget.inverse());
          Transform3d fieldToRobot = fieldToCamera.plus(robotToCamera.inverse());
          Pose3d robotPose = new Pose3d(fieldToRobot.getTranslation(), fieldToRobot.getRotation());
```

this code, for our single tag estimation, looks like it could be manipulated a bit (just by giving public access to more of these variables) in drive to produce a close estimate, still deals with ambiguity though due to target.bestCameraToTarget

linglejack06:
> @linglejack06 yeah I had already read the MA build blog when making my suggestion. In the blog they post links to their code for computing pose from single tag using a trig based 2d approach and I thought that would be worth looking into for crosstrack error.
> 
> 3d mode will always give a good distance estimate

ok, ill look into it for crosstrack error, it is just a little hard to follow due to their use of their own vision system, but ill try to figure it out. but we are fine with the 3d estimate for distance even though the heights will be very similar? I think the takeaway was if we do that we just have to make sure our angles and offsets are exact

linglejack06:
would the code shown above work for crosstrack error at all?

linglejack06:
is this code `Transform3d cameraToTarget = target.bestCameraToTarget.getTranslation().getNorm();` grabbing the 3d distance like implied in the post? want to double check before i go off of this distance

jkleiber:
> is this code `Transform3d cameraToTarget = target.bestCameraToTarget.getTranslation().getNorm();` grabbing the 3d distance like implied in the post? want to double check before i go off of this distance

@linglejack06 yeah that is my interpretation but it's weird that it's a Transform3d

aidnem:
Yeah it seems like getNorm returns a double according to WPIlib javadocs.

@linglejack06 After it calculates where it is vs. where it should be, will we just PID goal speeds to the right location? How do you intend to actually drive there?

jkleiber:
@aidnem my expectation is that we'd have something like the following control architecture:

**Forward Velocity Command**
distance_to_reef_trough = vision.getTagAlongTrackError() // implement this
fwd_motion_profile <-- generate this based on forward distance to go every cycle
v_forward = fwd_motion_profile(t=0)

This sort of turns into model predictive control (MPC), but isn't quite the same since we aren't really modeling anything. One concern I'd have with this approach is if it's jittery.

_Alternatively:_
v_forward = pid(distance_to_reef_trough) // aim to reduce this to 0.

PID gains would need to be tuned for short distances. Velocity also should be limited to a decently slow speed to avoid commanding an insane velocity

**Lateral Velocity Command**
xtk_error = vision.getTagCrossTrackError() // implement this
v_xtk = pid(xtk_error)

**Rotation Command**
heading_error = reef_heading - robot_heading
omega = pid(heading_error)

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -389,6 +428,149 @@ public Command getDriveToPoseCommand() {
...
+
+    DistanceToTag observation = alignmentSupplier.get(tagId, cameraIndex, 0, 0);
+
+    if (!observation.isValid()) {
+      return;
+    }
+
+    // give to PID Controllers and setGoalSpeeds (robotCentric)
+    double vx = driveLineupController.calculate(observation.alongTrackDistance());
+    double vy = driveLineupController.calculate(observation.crossTrackDistance());
```
jkleiber, line 609: You will want two separate PIDs here

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -389,6 +428,149 @@ public Command getDriveToPoseCommand() {
...
+    }
+  }
+
+  /**
+   * gets rotation for each side of hexagonal reef for lineup
+   * 
+   * @return Rotation2d representing desired rotation for lineup
+   */
+  public Rotation2d getRotationForReefSide () {
+    switch (desiredLocation) {
```
jkleiber, line 560: This switch/case should leverage case fall through to boil it down to 6 cases. For example:
```
switch(desiredLocation){
case Reef0:
case Reef1:
return new Rotation2d();
```
etc.

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -389,6 +428,149 @@ public Command getDriveToPoseCommand() {
...
+  }
+
+  /**
+   * gets tag to use for final alignment with vision
+   *
+   * @return int representing tag id to use
+   */
+  public int getTagIdForReef() {
+    boolean allianceRed = this.isAllianceRed();
+    switch (desiredLocation) {
```
jkleiber, line 488: Same comment as other switch statement, you can avoid repeating the same return statement by just stacking cases that are the same

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -389,6 +428,149 @@ public Command getDriveToPoseCommand() {
...
+        return -1;
+    }
+  }
+
+  /**
+   * gets camera index for vision single tag lineup
+   *
+   * @return 0 for Front Left camera; 1 for Front Right camera
+   */
+  public int getCameraIndexForLineup() {
```
jkleiber, line 523: Recommend making this an enum rather than an int so it's more readable (FrontRight, FrontLeft, Neither). Feel free to disregard if this makes life harder downstream though

linglejack06:
also, does my isOTFFinished command make sense? its checking if robot pose is close enough to desired pose by getting the norm between them. Should i instead check i both x and y are less than a value?

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -148,14 +150,22 @@ public enum DesiredLocation {
...
   private boolean isOTF = false;
 
+  private boolean isLiningUp = false;
+
   private Command driveToPose = null;
 
+  private VisionAlignment alignmentSupplier;
+
+  private PIDController driveLineupController = new PIDController(20, 0,0);
+  private PIDController rotationController = new PIDController(20, 0, 0);
```
jkleiber: We should JSON these gains at least, but having them in LoggedTunable would also be helpful for testing in shop in real time

jkleiber (commented):
First pass review

jkleiber:
@linglejack06 Using robot relative speeds is the correct technique for lining up to the reef tags since everything is from robot perspective

For the OTF question: norm is better because that's your actual distance away

src/main/java/frc/robot/subsystems/drive/Drive.java
linglejack06: ok, ill update that

src/main/java/frc/robot/subsystems/drive/Drive.java
linglejack06: i didnt know you could do that, can i do that for the getting camera index method as well? ie
``` 
case Reef0:
case Reef2:
case Reef4: ...
return 1;
case Reef1:...
return 0;
```

src/main/java/frc/robot/subsystems/drive/Drive.java
linglejack06: i could do this, we would again have to make an array to compare the enum to an int though, as vision wants a camera index as int, so im not sure if its worth it

src/main/java/frc/robot/subsystems/drive/Drive.java
jkleiber: yeah that's a good idea

src/main/java/frc/robot/subsystems/drive/Drive.java
jkleiber: probably not worth it then - instead, could you make some `final` ints that make it more obvious which camera index we are using:
```
private final int FRONT_LEFT_CAMERA_INDEX = 0;
private final int FRONT_RIGHT_CAMERA_INDEX = 1;
```

src/main/java/frc/robot/subsystems/drive/Drive.java
jkleiber: In test mode I'd like to have the loggedtunable for this at least. JSON for sure in all cases

jkleiber:
@linglejack06 by using `relativeTo` and `getNorm` you are computing the distance between robot and target position. This basically treats the robot as the origin

linglejack06:
> @linglejack06 by using `relativeTo` and `getNorm` you are computing the distance between robot and target position. This basically treats the robot as the origin

ok good, thats what i wanted just wanted to double check thats what was actually happening

linglejack06:
@jkleiber do you know why my getDistanceFromMultitag function isnt working? it just sends robot off field.

also, its hard to test results in sim, as for some reason simulated vision always sees like four tags (i think it thinks one of the cameras is behind robot) so whenever it goes to single tag result it almost immediately switches the single tag result to another id

jkleiber:
@linglejack06 idk why that function isn't working, but we can debug at shop. I think the first thing to figure out is which tags are in view and what cameras are seeing them. I think it may be helpful to log this in advantage scope by creating a set of lines that go from each camera to the tags they see and showing those on the field view (idk how hard this is)

linglejack06:
@jkleiber pid gains added as json and logged tunable. could you check implementation of logged tunable to see if im using it right?

also, drive is super long now lol

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -337,10 +509,78 @@ public void setOTF(boolean isOTF) {
...
+   * @return true if robot pose is close to desired pose
+   */
+  public boolean isDriveCloseToFinalLineupPose() {
+    // relative to transforms first pose into distance from desired pose
+    // then get distance between poses (if less than 0.1 meters we are good)
+    return this.getPose()
+            .relativeTo(this.findOTFPoseFromDesiredLocation())
+            .getTranslation()
+            .getNorm()
+        < 0.1;
```
jkleiber, line 529: We should make this JSON/tunable

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -337,10 +509,78 @@ public void setOTF(boolean isOTF) {
...
+   *
+   * @return true if location is reef; false otherwise (processor / coral station)
+   */
+  public boolean isDesiredLocationReef() {
+    if (desiredLocation == DesiredLocation.CoralStationLeft
+        || desiredLocation == DesiredLocation.CoralStationRight
+        || desiredLocation == DesiredLocation.Processor) {
+      return false;
+    }
+    return true;
```
jkleiber, line 563: Could we simplify this to: 
```
return desiredLocation == DesiredLocation.Reef
```

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -453,6 +693,140 @@ public Command getDriveToPoseCommand() {
...
+  }
+
+  /**
+   * gets rotation for each side of hexagonal reef for lineup
+   *
+   * @return Rotation2d representing desired rotation for lineup
+   */
+  public Rotation2d getRotationForReefSide() {
+    switch (desiredLocation) {
+      case Reef0:
```
jkleiber, line 771: Cases here can be combined

jkleiber (commented):
Second pass

I think you're using logged tunable correct, but you should try it in sim to confirm. If you tune for a really bad PID gainset it should be obvious that it "works" when the controller fails

src/main/java/frc/robot/subsystems/drive/Drive.java
linglejack06: ok ill add that as json

src/main/java/frc/robot/subsystems/drive/Drive.java
linglejack06: that would make is longer, since id have to check if it equals 0 or 1 or 2 or 3 or 4... or 11. we dont have just a DesiredLocation.Reef its all .Reef0 etc

src/main/java/frc/robot/subsystems/drive/Drive.java
linglejack06: ok ill add that as json

src/main/java/frc/robot/subsystems/drive/Drive.java
linglejack06: yes, just waiting to combine until i find proper rotations and poses that we want when i drive robot again

linglejack06: yes, just waiting to combine until i find proper rotations and poses that we want when i drive robot again

linglejack06: yes, just waiting to combine until i find proper rotations and poses that we want when i drive robot again

src/main/java/frc/robot/subsystems/drive/Drive.java
linglejack06: that would make is longer, since id have to check if it equals 0 or 1 or 2 or 3 or 4... or 11. we dont have just a DesiredLocation.Reef its all .Reef0 etc

src/main/java/frc/robot/subsystems/drive/Drive.java
linglejack06: ok ill add that as json

src/main/java/frc/robot/subsystems/drive/Drive.java
jkleiber: You could do

```
Return !(A || B || C)
```

Where the letters are the locations that aren't the reef

src/main/java/frc/robot/subsystems/drive/Drive.java
linglejack06: @jkleiber does rotation change depending on alliance, ie reef side closest to barge is 180 degrees, is that the same for blue/red or will one be 0 degrees

jkleiber: Blue and red will be offset by 180 since heading is field relative

linglejack06:
@jkleiber getting very close to working, as I left shop tonight I think the rotation was just wrong (tried to turn around then lost sight of tag).

linglejack06:
@aidnem if you get to test this today, there's a drive lineup tuning test mode, that has all the pid gains as logged tunable (along track meaning forward / reverse from tag, cross track meaning strafing, and rotation)

Also offsets can be added for the lineup, if drive isn't getting close enough or is too far away, it's in lineup method where I call alignmentsupplier.get

sleepyghost-zzz:
@linglejack06  are you going to be here today?

linglejack06:
No I won't be there

aidnem:
@jkleiber rotation was messed up because the rotation controller wasn't continuous, fixing that fixed the rotation problem.

We tried to implement a motion profile but it seemed to ramp up really slowly and generally be incredibly sketchy (needs more tuning probably, I only had like 20 minutes).

We've noticed that everything dealing with AlongTrack is negative (because robot relative +X is forward, while positive error also means we need to drive forward) so there are a lot of negative signs in the code, but after setting the AlongTrack offset to a negative number apparently it stopped roughly the correct distance away from the reef according to @avidraccoon (I had left by that point)

jkleiber:
@aidnem +X forward makes sense, but with positive X error I would expect +X output. We probably should log the controller output to ensure it is positive 

I'm not sure it's a great idea to make the offset negative, since that doesn't match reality. I think we ahould dig a little bit deeper to make everything a bit more intuitive

aidnem:
It's because we're passing it along track as error, which means its trying to control from a positive number down to zero, hence negative output (I think). We should figure out what we need to negate to make the code intuitive & match reality.

linglejack06:
@jkleiber @aidnem probably easiest to do is switch offset form addition to subtraction in vision method in coppercore, then you can pass positive offset in through drive

jkleiber:
@aidnem P control equation:

control = Kp * error

Since error > 0, control > 0 so I would expect the robot to go forward 

@linglejack06 When the offset was 0 we were driving backwards. I think making the offset a subtract makes sense but I'm not convinced it's the root cause

linglejack06:
Interesting, we can checks it out Monday, probably just a logic error or math error in vision

linglejack06:
@jkleiber might be because camera is upside down, so then tag angle is viewed off, and maybe the sine is reporting negative?

Because rotation 3d of camera is all 0 (that's the only way it would work without pose being flipped in some way on field) but since camera is still upside down it might be reading opposite? Does that make sense

jkleiber:
Along track distance can't be negative unless one of the angles is greater than 90 degrees (since that calculation only uses cosine)

linglejack06:
@jkleiber so could upside down camera cause greater than 90 degree angles? I wouldn't think so, maybe distance 3d is negative because we are going towards origin on field on red side? Ie higher x to lower x coordinates according to field

jkleiber:
@linglejack06 PhotonVision wasn't reporting high values for the angle - it was what we expected in pitch and roll. Also we had the right tag (id 7) so I don't think alliance color played a factor

After thinking it over, my POV is that the "front" of the drivetrain for test drivebase just happens to be backwards based on our calibration. Rather than making this too complicated I think we should do the following:
1. Make the offset use subtraction in coppercore
2. Allow PID gains to be negative 
3. Undo the change to negate PID control output
4. Set robot specific lineup gains in the JSON files, and use negative gains for this robot

@aidnem did you happen to push those changes to coppercore? I think we should do step 1 on that branch

linglejack06:
> @linglejack06 PhotonVision wasn't reporting high values for the angle - it was what we expected in pitch and roll. Also we had the right tag (id 7) so I don't think alliance color played a factor
> 
> After thinking it over, my POV is that the "front" of the drivetrain for test drivebase just happens to be backwards based on our calibration. Rather than making this too complicated I think we should do the following:
> 
> 1. Make the offset use subtraction in coppercore
> 2. Allow PID gains to be negative
> 3. Undo the change to negate PID control output
> 4. Set robot specific lineup gains in the JSON files, and use negative gains for this robot
> 
> @aidnem did you happen to push those changes to coppercore? I think we should do step 1 on that branch

@jkleiber 
1. changes were pushed to coppercore on vision-changes branch, (use subtraction) ready for review
2. pid gain setters fixed to allow negatives
3. wheres the negate pid control output change? im not seeing it?
4. json gains are already set up so we can just change in the json and it will work

jkleiber (approved):

PR #58, opened 2025-01-29
Adds Angle Aligning

sleepyghost-zzz:


src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -390,7 +392,18 @@ public Command getDriveToPoseCommand() {
   }
 
   /** Runs the drive at the desired speeds set in (@Link setGoalSpeeds) */
-  public void runVelocity() {
+  public void runVelocity(Rotation2d desiredRotation) {
+    if (isAngleAlign) {
+      ProfiledPIDController angleController =
```
jkleiber, line 397: This should probably be a class variable rather than a local variable. It is resetting every frame which is probably not helping your cause

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -390,7 +392,18 @@ public Command getDriveToPoseCommand() {
...
+          new ProfiledPIDController(5, 0.0, 0.4, new TrapezoidProfile.Constraints(8.0, 20));
+      angleController.enableContinuousInput(-Math.PI, Math.PI);
```
jkleiber, line 399: Recommend configuring this in the constructor

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -390,7 +392,18 @@ public Command getDriveToPoseCommand() {
...
+      double omega =
```
jkleiber, line 400: Recommend logging current rotation, desired rotation, and computed omega to help debug in advantage scope

jkleiber (commented):
@sleepyghost-zzz here are some initial ideas to help you debug

sleepyghost-zzz:
https://github.com/user-attachments/assets/95d84e91-d99e-4fd5-aea8-14e13171b7f1

proof that it actually does the thing. this is controlled via joystick buttons. There are no confilcts with OTF or steering and stuff. :D

src/main/java/frc/robot/InitBindings.java
```diff
@@ -58,7 +59,22 @@ public static void initDriveBindings(Drive drive) {
...
-    // pov right (reef 0-11 -> processor left -> processor right )
+    leftJoystick
+        .button(1)
+        .onTrue(
+            new InstantCommand(
+                () -> {
+                  drive.lockRotationToPosition(new Translation2d(3.5, 3));
+                },
+                drive));
+    leftJoystick
```
jkleiber, line 70: just for future reference, you can add the `.onFalse()` after the `.onTrue()` rather than re-referencing the button.

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -146,9 +148,16 @@ public enum DesiredLocation {
   private DesiredLocation desiredLocation = DesiredLocation.Reef0;
 
   private boolean isOTF = false;
-
   private Command driveToPose = null;
 
+  private final ProfiledPIDController angleController =
+      new ProfiledPIDController(5, 0.0, 0.4, new TrapezoidProfile.Constraints(1000, 2000));
```
jkleiber, line 154: Can you add these gains to a JSON so we can configure them more easily?

Also 1000 and 2000 seem like insane choices for constraints. I think we should probably pick some real values here

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -146,9 +148,16 @@ public enum DesiredLocation {
...
+
+  private boolean lockRotationToPoint;
+  private Translation2d lockedTargetPosition = new Translation2d();
+  private Translation2d redReefCenter = new Translation2d(4.5, 4);
+  private Translation2d blueReefCenter = new Translation2d(4.5, 4);
```
jkleiber, line 159: How do both reefs have the same center?

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -426,8 +445,35 @@ public Command getDriveToPoseCommand() {
...
+
+    // Use PID to rotate toward the target angle
+    omega = angleController.calculate(getRotation().getRadians(), targetAngle);
+    setGoalSpeeds(
+        new ChassisSpeeds(goalSpeeds.vxMetersPerSecond, goalSpeeds.vyMetersPerSecond, omega),
+        false);
+
+    if (DriverStation.getAlliance().isPresent()
+        && DriverStation.getAlliance().get() == Alliance.Red) lockRotationToPosition(redReefCenter);
+    else lockRotationToPosition(blueReefCenter);
```
jkleiber, line 468: please use curly braces even though you don't technically have to

also this should be in the `alignToFieldElement` function (renamed version of `lockRotationToPosition`) rather than calling it. This does not need to be called every cycle.

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -305,6 +314,16 @@ public void setGoalSpeeds(ChassisSpeeds speeds, boolean fieldCentric) {
     }
   }
 
+  public void lockRotationToPosition(Translation2d targetPosition) {
+    lockRotationToPoint = true;
+    lockedTargetPosition = targetPosition;
+    angleController.reset(getRotation().getRadians());
```
jkleiber, line 320: currently it seems that you are resetting the angleController every cycle. That's not ideal, so could we remove this?

src/main/java/frc/robot/InitBindings.java
```diff
@@ -58,7 +59,22 @@ public static void initDriveBindings(Drive drive) {
                 },
                 drive));
 
-    // pov right (reef 0-11 -> processor left -> processor right )
+    leftJoystick
+        .button(1)
+        .onTrue(
+            new InstantCommand(
+                () -> {
+                  drive.lockRotationToPosition(new Translation2d(3.5, 3));
```
jkleiber, line 67: We shouldn't be passing in a magic position like this - the drivetrain subsystem should handle this itself based on a position target (making an enum with the only value being Reef would be a good starting point)

jkleiber: we'll add to that as we go (coral station, net) so you could add those as well.

I would also like this function to be more named something more general than `lockRotationToPosition` - maybe `alignToFieldElement` would be good

src/main/java/frc/robot/subsystems/drive/Drive.java
jkleiber: this function should be where you set the target position based on the provided option from the command (based on the enum I mention in another comment, and which alliance we are on).

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -426,8 +445,35 @@ public Command getDriveToPoseCommand() {
     return AutoBuilder.pathfindToPose(targetPose, constraints, 0.0);
   }
 
+  public void rotationallyLockedIn() {
```
jkleiber, line 448: Could we rename this to `alignToTarget` or similar? As written, it is ambiguous as to whether or not it will return a boolean

src/main/java/frc/robot/subsystems/drive/DriveConstants.java
jkleiber, line 1: Can we make this a JSON capable file once and for all? This is one of the only constants file that isn't

jkleiber (commented):
First pass review

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -315,6 +330,15 @@ public void setGoalSpeeds(ChassisSpeeds speeds, boolean fieldCentric) {
     }
   }
 
+  public void alignToFieldElement(Translation2d targetPosition) {
```
jkleiber, line 333: Recommend using an enum here rather than Translation2d and then set a variable called something like `targetFieldElement` based on that enum

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -410,21 +434,59 @@ public void updateDesiredLocation(DesiredLocation location) {
    */
   public Pose2d findOTFPoseFromPathLocation() {
     switch (this.desiredLocation) {
-        // reef 0 and 1 will have the same path
-        // NOTE: use PathPlannerPath.getStartingHolonomicPose to find pose for reef lineup if wanted
       case Reef0:
         return DriverStation.getAlliance().orElse(Alliance.Blue) == Alliance.Red
-            ? JsonConstants.redFieldLocations.reef0
-            : new Pose2d();
+            ? new Pose2d(
```
jkleiber, line 439: thank you so much for actually filling out all these fields, this is extremely based

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -453,8 +515,31 @@ public Command getDriveToPoseCommand() {
...
+    double omega = 0.0;
+
+    // Get current position
+    Translation2d currentPosition = getPose().getTranslation();
+
+    // Calculate desired angle to face the target position
+    double targetAngle =
+        Math.atan2(
+            lockedAlignPosition.getY() - currentPosition.getY(),
+            lockedAlignPosition.getX() - currentPosition.getX());
```
jkleiber, line 528: This is where my comment about having an enum and my comment about moving the alliance logic will come together to make the most sense.

You've built all this good structure to get us aligned to the center of the reef. The only problem is that it's very bespoke (it only has the reef in mind). Eventually we want to do heading alignment to the coral station, the net/barge, and the processor as well. And none of those heading targets ever change (i.e. the net is always 0 or 180 degrees depending on alliance color)

So given that you have a `targetFieldElement` that could be `Reef`, `Barge, `Processor`, `LeftStation`, `RightStation` we eventually could do:
```
switch (targetFieldElement) {
  case Reef:
    // atan2 logic you already have
    break;
  case Processor:
    // alliance based heading
    break;
  case LeftStation:
    // alliance based heading
    break;
...
  default:
    // disable heading alignment due to invalid target
}
```

Not asking you to implement the other heading targets yet, but I'd like to ensure we have a scalable solution for these future elements

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -453,8 +515,31 @@ public Command getDriveToPoseCommand() {
...
+
+    // Use PID to rotate toward the target angle
+    omega = angleController.calculate(getRotation().getRadians(), targetAngle);
+    setGoalSpeeds(
+        new ChassisSpeeds(goalSpeeds.vxMetersPerSecond, goalSpeeds.vyMetersPerSecond, omega),
+        false);
```
jkleiber, line 534: I can't recall if `false` means robot centric or field centric. Can you confirm it is field centric?

jkleiber (commented):
another review pass

src/main/java/frc/robot/subsystems/drive/Drive.java
sleepyghost-zzz: false is robot centric, true is field centric, because it is just updating rotation, if you put field centric on field centric, it gets really broken, and inverts controls while locked on the reef.

jkleiber (approved):

PR #59, opened 2025-01-29
Add claw and transition ElevatorSubsystem to ScoringSubsystem

aidnem:
Adds a claw, adds Scoring state machine to handle state and hopefully make it easy to control.

Also adds ScoringSetpoints.java, a new constants file that keeps records of every single setpoint, which can be tuned through JSON and should make tuning and using the setpoints in code easy.


src/main/deploy/constants/FeatureFlags.json
jkleiber, line 1: I don't think this should be here - the platform specific files are where these should be added

src/main/deploy/constants/ScoringSetpoints.json
```diff
@@ -0,0 +1,68 @@
+{
+  "idle": {
+    "elevatorHeight": {
+      "value": 1.5,
+      "unit": "Inch"
+    }
+  },
+  "l1": {
```
jkleiber, line 8: nit: can we capitalize the L in this and the other ones

at first glance I thought this said 11, 12, 13, 14 which made no sense

src/main/deploy/constants/comp/ClawConstants.json
jkleiber, line 1: just a note that you'll probably need to add these new constants files to the environment file thing

src/main/java/frc/robot/constants/ElevatorConstants.java
```diff
@@ -170,6 +170,12 @@ public final class ElevatorConstants {
   public final Integer CRTticksPerRotation =
       4096; // CANCoders have a resolution of 4096 ticks/rotation
 
+  /**
+   * How many meters away from setpoint the elevator height can be while still being considered "at
+   * the setpoint"
+   */
+  public final double elevatorTargetThresholdMeters = 0.03;
```
jkleiber, line 177: might as well make this a Units right?

src/main/java/frc/robot/constants/ScoringSetpoints.java
```diff
@@ -0,0 +1,45 @@
...
+import coppercore.parameter_tools.json.JSONExclude;
+import coppercore.parameter_tools.json.JSONSync;
+import coppercore.parameter_tools.json.JSONSyncConfigBuilder;
+import edu.wpi.first.units.measure.Distance;
+import edu.wpi.first.wpilibj.Filesystem;
+
+public class ScoringSetpoints {
+  /** Keeps track of a setpoint for the scoring subsystem. */
+  public static record ScoringSetpoint(Distance elevatorHeight) {}
+  ; // TODO: Add wrist angle to ScoringSetpoint
```
jkleiber, line 15: delete random semicolon?

src/main/java/frc/robot/subsystems/scoring/ClawMechanism.java
```diff
@@ -0,0 +1,60 @@
...
+    io.applyOutputs(outputs);
+
+    Logger.processInputs("claw/inputs", inputs);
+    Logger.processInputs("claw/outputs", outputs);
+  }
+
+  /** This method must be called from the subsystem's test periodic! */
+  public void testPeriodic() {
+    switch (TestModeManager.getTestMode()) {
+      case ClawTuning:
```
jkleiber, line 48: it might be worth tuning the stopping time of the claw based on the CANRange as a loggedtunablenumber

Also it would be good to try out using velocity control for holding pieces (algae primarily). Idk if it'll work and it doesn't have to be now, just an idea

src/main/java/frc/robot/subsystems/scoring/ScoringSubsystem.java
```diff
@@ -0,0 +1,222 @@
...
+
+    stateMachineConfiguration = new StateMachineConfiguration<>();
+
+    stateMachineConfiguration
+        .configure(ScoringState.Idle)
+        .permit(ScoringTrigger.BeginIntake, ScoringState.Intake);
+
+    stateMachineConfiguration
+        .configure(ScoringState.Intake)
+        // .permitIf(ScoringStateMachineTriggers.BeginIntake, ScoringStateContainer.Idle, () ->
```
jkleiber, line 94: does this not work?

src/main/java/frc/robot/subsystems/scoring/states/ScoreState.java
```diff
@@ -0,0 +1,21 @@
...
+  private ScoringSubsystem scoringSubsystem;
+
+  public ScoreState(ScoringSubsystem scoringSubsystem) {
+    this.scoringSubsystem = scoringSubsystem;
+  }
+
+  @Override
+  public void periodic() {
+    scoringSubsystem.setElevatorGoalHeight(JsonConstants.scoringSetpoints.idle.elevatorHeight());
+    scoringSubsystem.setClawRollerVoltage(Volts.zero());
```
jkleiber, line 19: fwiw this does need to be updated to not be idle (although you probably know this)

src/main/java/frc/robot/subsystems/scoring/states/WarmupState.java
```diff
@@ -0,0 +1,53 @@
...
+      case Processor:
+        setpoint = JsonConstants.scoringSetpoints.processor;
+        break;
+      case Ground:
+      default:
+        System.out.println(
+            "ERROR: Can't warmup for FieldTarget "
+                + scoringSubsystem.getTarget()
+                + ", defaulting to L1");
+        setpoint = JsonConstants.scoringSetpoints.l1;
```
jkleiber, line 46: recommend going to idle if the warmup is impossible

jkleiber (commented):
First pass

idk if the code will work, but it looks great structurally. I think we won't be in immense pain trying to debug this which is nice

src/main/java/frc/robot/subsystems/scoring/ClawMechanism.java
aidnem: Do you mean like control rollers with closed loop to a velocity?
Also I'm not against it but what are you thinking it'll help us do?

jkleiber: It's an idea for holding the algae in the rollers without needing to apply a constant voltage

It might not work if the algae doesn't make the wheels spin when it's falling out, which is why it's just a thought. We should have a test mode for figuring out how to hold algae (i.e. configurable holding voltage or velocity PID)

If voltage works I don't really think we need to do velocity PID

aidnem: Okay sounds good.
I was also thinking, could we control stopping time by rotations instead of time? Theoretically wouldn't the motors spin a certain amount to move the coral a certain amount?
I'll probably also fully rewrite sim because I need to simulate the actual position of the object instead saying "if you power on for a full second you intake a piece"

src/main/java/frc/robot/subsystems/scoring/states/ScoreState.java
aidnem: We don't even really have to set this here, do we? The height should be set by WarmupState and left sitting during this state.

src/main/java/frc/robot/subsystems/scoring/ScoringSubsystem.java
aidnem: It should, Nathan was showing me something as another option and then I forgot to pick one and just left both with one commented out.

src/main/java/frc/robot/subsystems/scoring/states/ScoreState.java
aidnem: I could also generalize this into a helper in ScoringSetpoints.java to get the warmup setpoint and print error messages to avoid the duplicated code, I'll do this instead :)

aidnem:
@jkleiber We want to handle picking up an algae from the reef as intaking and not scoring right? This is more just me checking my thinking but we want the rollers to continously run and the mechanical stuff to move itself to the right position as soon as it can and everything.

jkleiber:
@aidnem that is correct. Algae intake has 3 positions (L2, L3, ground) and Algae Score has 2 positions (net, processor)

aidnem:
Awesome, thanks.

Okay to-do for this to get merged:

- [ ] Claw intaking test mode
- [ ] Lore-accurate claw-sim with actual moving object (probably using mechanism2d to visualize)

src/main/java/frc/robot/subsystems/scoring/ClawMechanism.java
jkleiber: yeah that's actually an inspired idea, let's try it

aidnem: FOC velocity control could be incredible for this because it would actively oppose the motion and bring it to a faster stop than just setting 0 volts and hoping the object stops.

aidnem:
Sim doesn't work quite right but the tuning mode does:

https://github.com/user-attachments/assets/2d883f07-ee1d-4e8e-a0a9-21166164b4f1

It's a little inconsistent at 3V and I suspect it's because of slow CAN refresh rate. Ideally we'd put a canivore on this bus too.
Observing the graph shows stairsteps of ~0.5 rotations where precision much more than that is preferable.

At 0.5 volts it's super slow but works like a charm:

https://github.com/user-attachments/assets/f53b688c-5cec-4328-9621-d5135b124294

aidnem:
Sim is still a little suspicious but the code works for coral on the HITL rig. I have to test algae intaking still. We were testing with keeping a motor stalled and that seemed to work better when holding an algae, so I'm considering adding that to the code as well.

Intaking was slightly inconsistent at 3V but I'm going to try it on a canivore today because the refresh rate for signals from the motor for position readings was super slow.

aidnem:
@jkleiber what's your opinion on merging this before getting it 100% done on hardware since they're still prototyping claws? I can make a new branch to integrate it whenever I need and this way I can start on wrist without making a huge PR that needs review.

jkleiber:
@aidnem I'm fine with merging before it goes to the hardware if it works reasonably well in sim. Just looking for first order correctness and we can iron out bugs later

aidnem:
Here's verification of sim working for coral:

image

You can see it waits until "diff" is at the threshold and then stops applying voltage.

Algae coming soon :)

aidnem:
And here's lore accurate algae:

image

aidnem:
Should be ready now, I assume there will probably be some review stuff that needs to be fixed but as a whole I feel like its ready.

src/main/java/frc/robot/subsystems/scoring/ClawIOSim.java
```diff
@@ -0,0 +1,176 @@
...
+      }
+    } else {
+      piecePos +=
+          RotationsPerSecond.of(
+                      outputVoltage.in(Volts)
+                          * JsonConstants.clawConstantsSim.rotationsPerSecondPerVolt)
+                  .times(Seconds.of(0.02))
+                  .in(Rotations)
+              * 0.05
+              * (has == HasState.ALGAE ? -1.0 : 1.0);
```
jkleiber, line 98: Not blocking, but why not just save the last position and compare it to the current position?

We can re-evaluate in a future PR if needed

jkleiber (approved):
auto-merge enabled - ship it

we can continue to improve this as we go but it seems sane for now from a structural perspective (and we can remove the debugging stuff later)

src/main/java/frc/robot/subsystems/scoring/ClawIOSim.java
aidnem: Sorry I'm not sure what you mean. If you're talking about the invert, it's because the claw wheels push an algae in the opposite direction as they push a coral.
If you elaborate a little I can fix it in the wrist PR (or decide if it needs fixing).

jkleiber: Is it possible to track the delta of the encoder position of whatever motor is running, or take the real velocity of the motor rather than doing rotationsPerSecondPerVolt?

aidnem: Oh this is just for sim, I'm trying to simulate it spinning faster with higher voltages.

jkleiber: Oh yeah, carry on then lol

aidnem: For the record it does actually use motor's encoder to count rotations irl don't worry

PR #70, opened 2025-02-04
Add States to drive

linglejack06:
- add lineup state to handle lineup logic, otf for otf, joystick for joystick
- add intakeLocation and goToIntake to help determine when we want coral station vs reef / processor for otf
-


linglejack06:
@jkleiber @aidnem how do i make a state transition to another state upon a condition being met? ie if cancel otf i want joystick state, or if otf completes then i want lineup

also, what should even go into JoystickState, should i just make drive with joysticks send its inputs there and then joystick state can setGoalSpeeds?

jkleiber:
@linglejack06 for the state transition question, I assume we could set a flag from a command that each state can either handle or not depending (for example, setting otf flag would trigger joystick state to attempt to transition to otf state. Likewise if otf flag goes false otf would fire the edge to go to joystick state)

As for the drive with joysticks, I would keep sending to the overall subsystem and pass those inputs into the joystick state I think

linglejack06:
@jkleiber so just set goal speeds in drive for joysticks, but from state or command?

jkleiber:
@linglejack06 if I understand correctly, the way it currently works is that drive with joysticks sends Drive some goal speeds and they get applied

So what we should do is pass those goal speeds to the relevant state(s) and have them apply it when they run their periodic. Otherwise they just get ignored

So short answer: in the states

linglejack06:
@jkleiber @avidraccoon how do i run the onEntry functions?

also, how can i re run on entry? for example, if we change location in the middle of otf, i want to rerun its onEntry (which schedules the otf command) so it can go to the new location. Do i have to fire state to cancel and then refire the otf state?

linglejack06:
also, can i cancel a state from within that state, ie if lineup is running but it cant find a tag then i want to go back to joysticks. this logic is in the lineup state though, not drive, so i cant fire a trigger

jkleiber:
@linglejack06 take a look at scoring subsystem for how to fire triggers in states. @aidnem has been passing the subsystem to the state constructors 4o do this

linglejack06:
yeah, i saw after looking further, do you mind checking over the states to
see if thats how you want it structured?

linglejack06:
> Also on the on entry stuff I think that gets configured as part of the transition. You can also run the action again via the run on entry function in the state superclass I think? I'd need to take a look on a real code editor to be sure
> 
> I'll take a look at the state structure later today

ok, ive temporarily configured this function by just running on entry whenever the pose configured is different from drives desired location pose.

linglejack06:
also, on entry doesnt seem to be running, my otf state only began working once i added state machine periodic to drive periodic and i called otfState.onEntry in otf state periodic

avidraccoon:
That is a known problem, and there is a branch on coppercore with an almost finished fix.

avidraccoon:
The branch is for a few broken parts of the state machine I could also fix the not being able to pass variable in that.

jkleiber (commented):
super high level feedback, hopefully this helps with the problems you're seeing

linglejack06:
@jkleiber @aidnem could someone help me figure out why state cant transition to Lineup, it doesnt work from OTFState or from Joystick State. even though all necessary booleans are passing 

(ie permit if for otf requires that drive is close to lineup pose and desired location is reef, which i confirmed both are true) and joystick just requires that drive is close to lineup pose if beginAutoALignment trigger is called.

linglejack06:
> @linglejack06 if the state machine fails to fire a trigger I think there is some sort of error code it should provide. Does it do that, and if so, what is the error?
> 
> Also dumb debugging idea: if you println inside of the if statement next to the otf trigger we can at least confirm the state machine is attempting to fire the trigger

ill try the println, there are no errors relating to state machine i dont think, i believe part of what nathan is doing is adding those errors

linglejack06:
and you mean println next to the finishOTF trigger right?

avidraccoon:
Currently, no errors are thrown, but you can get information about the last attempted transition.

linglejack06:
@jkleiber update: print ln statement runs inside of the conditional that fires finish otf trigger

Screenshot 2025-02-06 at 11 27 36 AM

linglejack06:
> Currently, no errors are thrown, but you can get information about the last attempted transition.

i tried this @avidraccoon but the transition only shows up for one loop, so its imposible to find and cant be printed to console

linglejack06:
@avidraccoon the transition is null when its attempting to move from OTF to Lineup

avidraccoon:
Can you show me your state machine configuration?

linglejack06:
@avidraccoon its in drive.java, or do you need a screenshot

avidraccoon:
Found the problem I had a logic error in the state machine that made it so if you had more than one conditional transition that used the same trigger it would not work.

linglejack06:
@avidraccoon could it be that its constantly firing the trigger and never actually letting the state change, since its firing every loop

avidraccoon:
For now you can use maven local on the quickfix-statemachine branch of coppercore I believe it is up to date with main.

linglejack06:
> For now you can use maven local on the quickfix-statemachine branch of coppercore I believe it is up to date with main.

okay, could you make a pr with that change and publish a release too so we can test it easier at shop tonight

avidraccoon:
Also, for the problem, a conditional transition after the first one must have been evaluated to be true.

linglejack06:
> Also, for the problem, a conditional transition after the first one must have been evaluated to be true.

what do you mean?

avidraccoon:
Because you defined the transition to lineup second is the only reason that this happened between these two states

avidraccoon:
I have made a fix on the quickfix-statemachine branch, please let me know if it works or not

avidraccoon:
It also fixes the onEntry problem

linglejack06:
does on exit work too @avidraccoon

avidraccoon:
I believe because I have no evidence to say they don't and I fixed the problem onEntry had also for onExit.

linglejack06:
@jkleiber state works now, but sometimes it doesn't like going straight to lineup due to thinking it's not close enough to otf pose. Ie if manually driving close and then clicking, it tries otf because the pose is not close enough to final pose (it's closer) can you think of any logic to change this behavior? If I increase distance pose can be from final pose it doesn't help

jkleiber:
@linglejack06 I'm confused about the issue. If you're close enough to the OTF then lineup should happen automatically. Are you saying that the robot is closer to the reef than the OTF pose is so it goes backwards?

We could use the field finder utils in coppercore to see if you're already "on the runway" by seeing if you're in a certain rectangle. You'd need to define those rectangles but it could be a good add to skip OTF

linglejack06:
@jkleiber update: found logic that works by just checking if driveToPose command is finished, if so we fire beginLineup trigger.

Drive now transitions through all states, and can even reroute otf if changing desired location mid state

I think this should be ready for review and merge once I push

linglejack06:
https://github.com/user-attachments/assets/57377905-b95a-4652-b30f-323517b5f989

linglejack06:
Hold on review actually, I need to figure out why josh rotation isn't working anymore

linglejack06:
@jkleiber ready for review

- rotation fixed
- states tested
- added goToIntake variable and intakeLocation variable to drive (this way we can have a scoring and intake location set) both interface with OTFState, currently set up so trigger goes to reef, top button goes to coral station.
- this quick cycle has been tested on robot and works great (station pose just a little off)

linglejack06:
also, build failing due to need of new coppercore release

src/main/java/frc/robot/constants/field/RedFieldLocations.java
```diff
@@ -17,7 +17,8 @@ public class RedFieldLocations {
               .toString(),
           new JSONSyncConfigBuilder().setPrettyPrinting(true).build());
 
-  public Translation2d redReefCenterTranslation = new Translation2d(176.75, (158.5));
+  public Translation2d redReefCenterTranslation =
+      new Translation2d(Units.inchesToMeters(176.75), Units.inchesToMeters(158.5));
```
jkleiber, line 21: Let's get this updated to the correct values

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -131,7 +131,7 @@ public void testPeriodic() {
       scoringSubsystem.testPeriodic();
     }
     if (FeatureFlags.synced.getObject().runDrive) {
-      drive.testPeriodic();
+      // TODO: run lineup state test periodic
```
jkleiber, line 134: Can we add this back real quick

jkleiber (commented):
just some quick things to knock out before we merge this.

Also please update coppercore when able. I created a ticket for follow on improvements to the state machine

src/main/java/frc/robot/RobotContainer.java
linglejack06: @jkleiber i can't call a states method from robot container or drive I don't think, unless you know a way.

That's why it's currently called in the states periodic method IF driver station test is enabled, which should work the same. Just need to remove the to do from there

src/main/java/frc/robot/constants/field/RedFieldLocations.java
linglejack06: updated

jkleiber (approved):

linglejack06:
release published on coppercore, should fix build soon

src/main/java/frc/robot/RobotContainer.java
aidnem: @linglejack06 

You could probably access the state through the enum, since enums are static if I recall correctly:

```java
   Drive.DriveState.Idle.getState()./* do whatever you wanted to do here*/
```

Issue #67, opened 2025-02-04
Tune lineup controller for fastness


linglejack06:
**Purpose**
The purpose of this addition is to make the final lineup with the tag happen faster

**Project Scope**
- [ ] try using motion profile for alongtrack alignment
- [ ] tune gains in test mode
- [ ] add the encoder offsets, module positions, and other constant fields from test_drivebase  to comp drivetrain json too

jkleiber:
@linglejack06 before we commit to recomputing a motion profile every cycle to make this faster, I think we could try the following:
```
double maxVelocity = 3.0; // this is already set in constants somewhere
double maxAcceleration = 2.0; // this is already set in constants somewhere


// use vf^2 = vi^2 - 2ax to compute vi -> vi^2 = vf^2 + 2ax, vf^2 = 0 so vi^2 = 2ax
// x: alongTrackDistance
// a: maxAcceleration
// This assumes we are always decelerating. Since we have a max velocity we can clamp the result for vi
// to that value until we get below max velocity (when the equation tells us we need to decelerate)
double velocityCommandRaw = Math.sqrt(2*maxAcceleration*alongTrackDistance);
double clampedVelocity = Math.min(velocityCommandRaw, maxVelocity); // this is what we command
```
If we are at a slow speed and far away, we would send maxVelocity to the robot, but based on the current parameters that would not happen unless we are over 2.25 meters away. 

If we start at 0 speed we would always remain at or below the speed profile produced here and would have no issues, but it would be slow.

If we start above the recommended speed by this profile we could theoretically have issues (say we're going 3 m/s initially and then this says we should only be at 1 m/s because we are 0.25 meters away, we would crash into the reef pretty hard). To mitigate this, we should improve the OTF handoff logic some

The OTF handoff logic needs improvement anyway because waiting to get to the exact point for a reef target slows us down. We should check instead that the desired tag is in view and we are slower than (or close enough to if above) the speed target recommended above for lineup. That way we can transition with speed into lineup

linglejack06:
@jkleiber this could make sense, not sure how we would implement the otf handoff logic though (would i just let otf leave otf once vision has a distance error relative to its tag?

The issue i see with that is that all of that logic is centered in the lineup state, so this would require copying over those functions into otf.

Another option is we can just set a goal end velocity of 0.5/1 on otf path (this prevents robot from fully stopping then speeding up again for lineup) 

[...]

PR #72, opened 2025-02-07
add json configurable autos

linglejack06:
- add AutoPath.java for json sync


linglejack06:
@avidraccoon can i use lists for json sync or do i need to use an array?

avidraccoon:
@linglejack06 I believe either should work

avidraccoon:
For the array it might need to not be final

linglejack06:
going with list for now, ill see if it works

linglejack06:
@jkleiber two questions:
1. how should i go about giving different json templates to the AutoPath.java file? like how can i programmatically change the desired file name? do i just need different java files?
2. can you think of a better way to check if drive has finished autoAlignment? right now its just `return stateMachine.getCurrentState().equals(DriveState.Joysticks)` as i can assume in auto that once we enter joysticks, on the fly and lineup has finished

linglejack06:
i could make it so where it enters idle once wait on score or wait on intake (yet to be added) is false. then the auto command just waits for state to enter idle, once it does it refires auto alignment trigger

linglejack06:
@aidenm id also welcome your input for isDriveAlignment finished method and overall structure of AutoCommand if you want to take a look

jkleiber:
@linglejack06 There are couple perspectives I want to satisfy here (also FYI @aidnem):

1. Making a sane auto selector
2. Looking ahead to how we will want to do this in teleop (see discussion in https://github.com/team401/SnakeScreen/issues/17)

**First, sane auto selector:**
The role of JSON here is to allow for us to define the auto plan without recompiling. So we should create a class `AutoStrategy` that can load from JSON. JSON is always loaded at init time to avoid slowing down the controllers in a match.

But, we will want to be able to choose from various `AutoStrategy`s at runtime. Therefore, we should create an arraylist of `AutoStrategy` and basically load all the available auto strategies from a particular directory (which itself may or may not be configured by JSON in a different place). Then we can have an auto chooser that simply shows all the available AutoStrategy options and the driver can choose in elastic.

**But what about actually running what we've selected**
I think we should make a queue of actions that is maintained by a StrategySubsystem. On autonomous init, this subsystem would add the selected AutoStrategy to its work queue and then start executing the strategy. 

In full auto mode this would work by taking the active action from the top of the queue and scheduling a command with the command scheduler. In mixed autonomy mode this would ingest the existing bindings using setter methods (since it would be waiting for driver input). In manual autonomy mode, this subsystem would not be active.

Once the strategy subsystem detects that we are in the correct configuration to continue, it would fire the next command in the queue

**So then, commands**
what if we made a command for AutoScore and a command for AutoIntake? For now they would be simple, but could be scheduled via the StrategySubsystem. These commands would be always running and only exit due to certain criteria

**AutoScore**
Exit criteria: drivetrain in position (eventually: game piece no longer detected)

**AutoIntake**
Exit criteria: drivetrain in position (eventually: game piece in the robot)

**Why not sequential/parallel command group?**
I fear that doing this would make it hard to actually define the commands in JSON. Additionally I'm looking at teleop where we will want to give the robot potentially multiple actions in a row to execute and thinking this would be a more general solution.

Also this type of setup could be used in future seasons since it is not tightly coupled to the commands, but rather acts as a mission manager of sorts.

We can talk about this at shop tonight since I'll be there to help the rookie team. If any of this sounds insane lmk because I definitely could be missing something with how WPILib would interact with this

linglejack06:
@jkleiber i like the general structure of this idea, I'll take it and run with it to try to come up with something we can move forward with.

check out the auto score and auto intake command to see if they are something you like. These just take drive and score subsystem for now, with a location to go to. so they can be easily configured with json. Also, these would work with sequential command groups just by looping through all locations and setting them with a new command each time

```
List<Commands> linkedList
for (location) {
 linkedList.add(new AutoScore(drive, score, location);
}

// somehow destructure list into sequential command group 
```

or without it i guess we would just do the foillowing in periodic of Strategy subsystem
```
if(currentCommand.isFinished()) {
 new AutoScore(drive, score, scoringLocations.get(nextIndex).schedule()
```

linglejack06:
my one question is, do we actually want this strategy class to extend subsystem, or could it just be a generic class that we then call its periodic method in robotContainer periodic?

linglejack06:
@jkleiber i think we might need a strategy command rather than subsystem, as in autonomous, robotContainer is supposed to send a command to run at the beginning of auto, which could be an overarching strategy command. Im just not sure if commands can be scheduled from within a command.

jkleiber:
@linglejack06 I don't think we actually have to do that though. I'm pretty sure Robot.java just schedules the command that we give it from robot container. We could eliminate that interaction and then let our subsystem take over probably.

linglejack06:
> @linglejack06 I don't think we actually have to do that though. I'm pretty sure Robot.java just schedules the command that we give it from robot container. We could eliminate that interaction and then let our subsystem take over probably.

okay, ill look into it, @jkleiber could you check my logic in AutoCommand execute, if its good logic i can jsut copy it over to our strategy subsytem periodic. its a little confusing but i think it should work

aidnem:
@jkleiber tiniest nitpick available, would Strategy coordinator be a subsystem? I'd prefer to call it something like StrategyManager to emphasize that it isn't controlling hardware.

I kinda prefer just running the commands back to back ourselves because then we have more control (you could update the action queue while it's doing an action without having to reschedule the entire sequential group).

linglejack06:
> @jkleiber tiniest nitpick available, would Strategy coordinator be a subsystem? I'd prefer to call it something like StrategyManager to emphasize that it isn't controlling hardware.
> 
> I kinda prefer just running the commands back to back ourselves because then we have more control (you could update the action queue while it's doing an action without having to reschedule the entire sequential group).

im good with that regarding sequential command

jkleiber:
@aidnem I just chose strategy subsystem as a name to show that I want to run this in periodic when enabled. It doesn't even have to extend Subsystem if we don't want it to.

Also, if we are going for a more creative name, we could call it SnakeCharmer

jkleiber:
@linglejack06 I think the logic in that class is in the right direction. We'll just need to make it a bit more general when we add the queueing stuff. For example, I think at first we will just run whatever command is next in the queue once the active one is done:

```
Command activeCommand = null;
Queue<Action> actionQueue;

void SetAndScheduleActiveCommand() {

  Action action = actionQueue.pop();

  // build active command based on action

  CommandScheduler.schedule(activeCommand);
}

void periodic() {
  if (activeCommand == null) {
    SetAndScheduleActiveCommand();
  } else if (activeCommand isn't scheduled) {
    SetAndScheduleActiveCommand();
  }
}
```

linglejack06:
> @linglejack06 I think the logic in that class is in the right direction. We'll just need to make it a bit more general when we add the queueing stuff. For example, I think at first we will just run whatever command is next in the queue once the active one is done:
> 
> ```
> Command activeCommand = null;
> Queue<Action> actionQueue;
> 
> void SetAndScheduleActiveCommand() {
> 
>   Action action = actionQueue.pop();
> 
>   // build active command based on action
> 
>   CommandScheduler.schedule(activeCommand);
> }
> 
> void periodic() {
>   if (activeCommand == null) {
>     SetAndScheduleActiveCommand();
>   } else if (activeCommand isn't scheduled) {
>     SetAndScheduleActiveCommand();
>   }
> }
> ```

ok, where do we add actions then? within this same class? @jkleiber

aidnem:
Are we thinking an Action will be a record? (like an enum for intake or score, then a location?)

linglejack06:
> Are we thinking an Action will be a record? (like an enum for intake or score, then a location?)

this is what i have assumed, and is implemented in StateManager, ill move logic over later today @jkleiber @aidnem

aidnem:
Sounds good

jkleiber:
@linglejack06 @aidnem I am thinking "Action" is a JSON representable thing / snake screen inputtable thing. I'm just trying to create the concept of a standard message packet that can describe all actions we want to do.

How we implement that I'll leave up to y'all. It can be a record if that works easily (we'd just load in the subfields of a JSON object into a record / take the snake screen input as a record I imagine?), or a JSONSync class / typeadapter thing.

Examples:
coral score
```
type: Score
piece: Coral
location: Reef0
level: 4
```
algae score:
```
type: Score
piece: Algae
location: Net
level: n/a
```
intake
```
type: Intake
piece: Coral
location: CoralRight
level: n/a
```

aidnem:
Looks good to me, everything there can be represented by an enum and should be serializable (the ScoringSetpoint enum serializes fine)

linglejack06:
> @linglejack06 @aidnem I am thinking "Action" is a JSON representable thing / snake screen inputtable thing. I'm just trying to create the concept of a standard message packet that can describe all actions we want to do.
> 
> 
> 
> How we implement that I'll leave up to y'all. It can be a record if that works easily (we'd just load in the subfields of a JSON object into a record / take the snake screen input as a record I imagine?), or a JSONSync class / typeadapter thing.
> 
> 
> 
> Examples:
> 
> coral score
> 
> ```
> 
> type: Score
> 
> piece: Coral
> 
> location: Reef0
> 
> level: 4
> 
> ```
> 
> algae score:
> 
> ```
> 
> type: Score
> 
> piece: Algae
> 
> location: Net
> 
> level: n/a
> 
> ```
> 
> intake
> 
> ```
> 
> type: Intake
> 
> piece: Coral
> 
> location: CoralRight
> 
> level: n/a
> 
> ```

@jkleiber this is similar to what I have implemented now, my plan was to have a method in state manager that inputs a json path and will convert those into actions.

In addition, snakescreen can easily be done a similar way, just grabbing relevant values from the network tables. This does not require any additional work on snake screen. Just getting values from network tables in periodic of state manager (similar to how it's done now in drive to get reef location)

linglejack06:
Snakescreen could also maybe be configured to send a whole action, but this just seems like an unecessary step to me, when network tables are configured how they are and can be implemented easily

linglejack06:
@jkleiber check out the current implementation, i think its fully ready for auto.

went ahead and added modes to the strategy manager (full teleop and manual) in manual i assume we will just not schedule commands for drive and schedule different commands for scoring, you should be able to see how it works though with the switch statement, etc.

linglejack06:
auto has been fully updated to run with strategy manager, teleop can easily be updated by just replacing the instant commands that set relevant subsystems to instant commands that add an action to strategy manager

linglejack06:
i think we should be able to add a json file relevant to the ExampleAutoPath and have autos running tomorrow for the meeting

src/main/java/frc/robot/commands/AutoCommand.java
jkleiber, line 1: is this file relevant anymore?

src/main/java/frc/robot/constants/autoStrategies/ExampleAutoPath.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.List;
+
+public class ExampleAutoPath implements AutoStrategy {
+  @JSONExclude public static final String autoStrategyName = "ExampleAutoPath";
+  public static final JSONSync<ExampleAutoPath> synced =
+      new JSONSync<ExampleAutoPath>(
+          new ExampleAutoPath(),
+          Filesystem.getDeployDirectory()
+              .toPath()
+              .resolve("constants/autoStrategies" + autoStrategyName + "/.json")
```
jkleiber, line 19: probably don't want the "/' in "/.json" here

jkleiber (commented):
looks decent enough to test although it would be good to see in sim first

linglejack06:
@jkleiber atrategy manager no longer has null errors, but for some reason the commands are immediately getting cancelled and it just progresses through the actions after one loop, I tried to mitigate this by just making command is finished return false, which made the progression issue go away, but still not commanding drive to go anywhere

jkleiber:
@linglejack06 when I test this in sim, I get the robot going to the first location correctly but then it just stays there forever

linglejack06:
> @linglejack06 when I test this in sim, I get the robot going to the first location correctly but then it just stays there forever

@jkleiber this was becasue command was always returning false for is finished, this was done to debug but im trying to run with real isFinished now.

jkleiber (approved):
@linglejack06 approved

We definitely need to prioritize #67 now because that's the limiting factor

PR #73, opened 2025-02-07
Add wrist and protection logic

aidnem:
- Adds a WristMechanism to the ScoringSubsystem
- Adds protection logic to prevent claw from hitting crossbar


src/main/java/frc/robot/subsystems/scoring/ScoringSubsystem.java
```diff
@@ -20,8 +20,69 @@
...
+ *       elevator moves past 4) intake swings back into place
+ * </ul>
+ *
+ * <p>This multi-clamp system can pose an interesting challenge: how do we choose which clamp to
+ * apply? The correct way to solve this is to structure the if-statements in a way that the most
+ * restrictive clamps are checked first, so that when those clamps aren't set, the less restrictive
+ * clamps are checked. After all of the clamp conditions are checked, we can finally put a call to
+ * reset the clamps in an 'else,' allowing a full range of motion when no protection conditions are
+ * met.
+ */
```
aidnem, line 82: @jkleiber would you mind looking over this logic and seeing if it makes sense to you before I implement?

jkleiber: It makes sense to the extent that it can make sense given the complexity of the problem.

I don't know what the crossbar is - is that the intake funnel?

The way to make this code the cleanest is to probably have a helper function for each mechanism that can do a collision: `DetermineElevatorClamp`, `DetermineGroundIntakeClamp` etc.

Then I would structure this as follows:
1. Get raw (unclamped) setpoints
2. Based on current positions and raw setpoints, determine clamps
3. Clamp the setpoints

That way you aren't cooked by stuff like which subsystem gets clamped first, and there would be a clear prioritization within each helper function (i.e. the ground intake wants to go in, but the elevator wants to go up AND is currently down -> we have to move the ground intake since elevator has priority for that specific maneuver)

jkleiber: It would be a good idea to make a spreadsheet or doc with all the scenarios and our expected response. You could just make a markdown file in the repo tbh

aidnem: I'll make that. Also I was informed yesterday that ground intake is the most likely piece not to be ready/be cut if we need to lose weight, so I think I'll save the protection logic for that for another ticket once the ground intake is implemented.

aidnem: @jkleiber I added [this markdown file](https://github.com/team401/2025-Robot-Code/pull/73/commits/0e9ea17f3fa1240610d68745df428963f0010655?short_path=3dcfe63#diff-3dcfe63e4ca674d99aa397cdc8c6f018c29f808185eb0cd37050db8de749ef94). It isn't exhaustive but I *think* it covers all of the cases where collisions could happen, at least without ground intake.

jkleiber: nice that looks good

aidnem:
@jkleiber The vibes I'm getting from the sim are that perhaps the zero in CTRE closed loop and the zero in the single jointed arm sim aren't the same value. Something must be messed up because its somehow oscillating when using only kG, which should either send it all the way up, all the way down, or hold position right? If kG is being calculated correctly I'm unsure how it could oscillate. Please let me know if I'm misunderstanding something, I've got it almost tuned with PID at this point but something seems off to me.

aidnem:
Oh also I've discovered an interesting behavior where it seems like if you drive the arm sim into its maximum or minimum angle too hard it takes a lot of force to move it off of the simulated "hardstops" again.

aidnem:
Changing to use voltage control behaved more as expected, I was able to tune it so that it goes to position with minimal overshoot.

jkleiber:
@aidnem that graph looks good. Why is the setpoint jumping up to the goal at the end?

I'm fine using voltage control here. We can switch to FOC if we have time to debug the apparent issues with it and feedforward

aidnem:
Thats something I wondered too. It seems to be inherent to MotionMagicExpo because it happened on the elevator too, but the range of motion was bigger so the graph was more zoomed out and the jumps were smaller.

It looks like its stopping short of the setpoint and then as soon as the setpoint slope goes to zero it snaps. I've used my epic art skills to extend the setpoint graph, it looks pretty horizontal:

2025_02_12_0gz_Kleki

aidnem:
Realized state machine for scoring was slightly outdated and have spent the last 2 hours rewriting it a little. Now I can test in sim.

aidnem:
image

Here's it almost working in sim. There is one slight issue, you can see where the clamp slightly jumps up, because it can't move down due to where wrist is, but then elevator overshoots its setpoint and since it thinks its safe it continues. This would probably smash the claw on the crossbar were this real life.

I need to debug why wrist wasn't clamped out of that scenario in the first place and also make sure setpoints are applied properly.

aidnem:
https://github.com/user-attachments/assets/53fabe56-1c5b-4a39-8413-8b270701da9e

Here's that same simulation in video form. It's basically just running cycles autonomously, except drive is featureflagged out so it doesn't have to wait for OTF & lineup.

You can see:
1. Wrist and elevator go to intake setpoint
2. After simulated claw intakes the piece, it continues to "warmup"
3. You can see the orange line, the clamped elevator goal height, wait for the wrist to be below a certain threshold before allowing elevator to go all the way up
4. Once it is close to setpoint, state changes to 'Score', and it quickly ejects the piece.
5. It then goes back to idle and then to intake and the process repeats.

aidnem:
Upon further inspection of the graph, I've discovered that the wrist overshoots its setpoint by 0.009 meters, causing it to be above the "maximum cross-bar-safe wrist angle" because my filler value for that setpoint is also the value for the clamp. This causes the elevator to briefly try to stop at 1.1 meters instead of 0.1 meters. However, the elevator overshoots its setpoint as well, causing that clamp to no longer be an issue (the elevator is now below the crossbar, so no worries anymore). This problem will be mitigated by better tuning and having actual setpoints that have a little bit of wiggle room in them.

image

Above is a close-up of that on the graph.

Yellow line is the elevator height, while orange line is its clamped goal position (observe the clamp coming up and the overshoot).

Blue line is the wrist position, and green line is the wrist "max angle", you can see that it's at zero, but the wrist angle is at ~0.01, causing that clamp to trigger. You can also see that once the elevator is fully past the crossbar, the wrist max angle goes back up, allowing it to move in under the crossbar (e.g. for coral station intake).

aidnem:
@jkleiber I *think* this is ready for another round of review. Let me know if you find any nitpicks, I have today off from school and I'd rather have this code immaculate than merge it in while it's bad.

jkleiber:
@aidnem taking a look now. It would be cool to see if this integrates well with the auto strategy stuff so could you take a look at that in the meantime (just trying to see if the correct signals are being passed to the right places, and that the scoring subsystem actually scores when lined up)

If it's super slow we can defer that testing because the lineup faster stuff is on a different branch

aidnem:
I'm testing autos in sim right now, a few things were broken and code is crashing but I'm not exactly stuck so I'll see how it goes.

aidnem:
Got autos to work, I have to switch to my desktop to screen record

aidnem:
https://github.com/user-attachments/assets/1d7576eb-fd19-4d5a-a621-abb3699b2005

@jkleiber here's an auto working in sim

jkleiber (commented):
some initial feedback

src/main/java/frc/robot/subsystems/scoring/ScoringSubsystem.java
```diff
@@ -227,7 +359,13 @@ public boolean isAlgaeDetected() {
...
    */
   public void setTarget(FieldTarget target) {
-    currentTarget = target;
+    // TODO: find and eliminate cases where null is passed to this function
+    if (target == null) {
+      System.out.println("WARNING: Null target set in setTarget");
+      new Exception("Null target in setTarget").printStackTrace();
+    } else {
+      currentTarget = target;
+    }
```
jkleiber, line 368: why not just do `if (target != null)` otherwise don't set

src/main/java/frc/robot/subsystems/drive/states/LineupState.java
```diff
@@ -88,6 +90,8 @@ public LineupState(Drive drive) {
   public void onEntry(Transition transition) {
     // cancel rotation lock on center
     drive.disableAlign();
+
+    ScoringSubsystem.getInstance().fireTrigger(ScoringTrigger.StartWarmup);
```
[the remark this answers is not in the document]
aidnem: Don't think it causes any problems. if we don't make it a singleton, I have to somehow pass it to all of the states that depend on it.

jkleiber:
Maybe things changed in your most recent commit @aidnem but in the video the elevator was only going to L1 (not asking for a new recording, things look good. Just trying to make sure we aren't missing anything there)

src/main/java/frc/robot/subsystems/drive/states/LineupState.java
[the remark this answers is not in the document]
jkleiber: if it works it works, I think this makes sense

src/main/java/frc/robot/StrategyManager.java
```diff
@@ -137,15 +137,16 @@ public Command getCommandFromAction(Action action) {
...
+              drive, scoringSubsysttem, action.location(), action.scoringTarget());
       }
     } else if (action.type() == ActionType.Score) {
       switch (this.autonomyMode) {
         case Full:
         case Teleop:
         case Manual:
         default:
-          return new AutoScore(drive, scoringSubsysttem, action.location());
+          return new AutoScore(drive, scoringSubsysttem, action.location(), action.scoringTarget());
```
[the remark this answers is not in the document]
aidnem: Done, I was ignoring that while fixing logic and then forgot to go back and change

src/main/java/frc/robot/subsystems/scoring/ScoringSubsystem.java
```diff
@@ -270,35 +402,53 @@ public void setOverrideStateMachine(boolean doOverrideStateMachine) {
...
   /**
-   * checks if other subsytems need to wait on intake
+   * checks if other subsystems need to wait on intake
    *
-   * @return true if intake has coral / algae
+   * @return false if intake has coral / algae, and true if its still waiting to intake
    */
   public boolean shouldWaitOnIntake() {
-    return false;
+    return !clawMechanism.isCoralDetected() && !clawMechanism.isCoralDetected();
```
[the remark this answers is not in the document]
aidnem: oops, one of those was meant to be algae

src/main/java/frc/robot/subsystems/scoring/ScoringSubsystem.java
aidnem: I wanted a stack trace in case I went back later and figured out why it was getting null sometimes and that null was overriding the actual correct FieldTarget that was being sent.

aidnem: I reordered the logic to indicate that we want it to not be null, otherwise we print a warning instead of vice versa

aidnem:
@jkleiber I realized there were some issues with multiple actions back to back to back and fixed those. I know you didn't ask for a new recording but I recorded one anyway because I didn't see that lol


https://github.com/user-attachments/assets/fa56c1c5-9e6d-4b12-9169-0c729f995eca

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -94,6 +91,15 @@ public void configureSubsystems() {
...
     }
 
+    if (FeatureFlags.synced.getObject().runScoring) {
+      scoringSubsystem = InitSubsystems.initScoringSubsystem();
+      if (FeatureFlags.synced.getObject().runDrive) {
+        scoringSubsystem.setIsDriveLinedUpSupplier(() -> drive.isDriveCloseToFinalLineupPose());
+      } else {
+        scoringSubsystem.setIsDriveLinedUpSupplier(() -> true);
+      }
+    }
```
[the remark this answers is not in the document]
aidnem: Created #80

src/main/java/frc/robot/constants/ScoringSetpoints.java
```diff
@@ -27,23 +29,38 @@ public static record ScoringSetpoint(Distance elevatorHeight) {}
...
+          "IdleWithAlgae",
+          Meters.of(0.2),
+          Rotations.of(-0.75)); // Hold algae with the wrist out and the arm up
 
   // TODO: Get actual values for all values from here down
-  public final ScoringSetpoint L1 = new ScoringSetpoint(Meters.of(0.05));
-  public final ScoringSetpoint L2 = new ScoringSetpoint(Meters.of(0.25));
-  public final ScoringSetpoint L3 = new ScoringSetpoint(Meters.of(0.5));
-  public final ScoringSetpoint L4 = new ScoringSetpoint(Meters.of(0.75));
+  public final ScoringSetpoint L1 = new ScoringSetpoint("L1", Meters.of(1.0), Rotations.of(0.0));
```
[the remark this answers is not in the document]
aidnem: Created #81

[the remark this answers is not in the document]
aidnem: Created #82  for setpoint determination

src/main/java/frc/robot/constants/subsystems/ElevatorConstants.java
```diff
@@ -166,6 +166,27 @@ public final class ElevatorConstants {
    */
   public final Distance elevatorTargetThresholdMeters = Meters.of(0.03);
 
+  /**
+   * This is the maximum height the elevator can reach while the wrist is "in" without smashing it
+   * into the crossbar.
+   */
+  public final Distance maxWristInBelowCrossBarHeight = Meters.of(1.0); // TODO: Real value
```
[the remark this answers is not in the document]
aidnem: #82 created

src/main/java/frc/robot/commands/strategies/AutoIntake.java
```diff
@@ -25,6 +32,20 @@ public void initialize() {
...
+      scoringSubsystem.setGamePiece(GamePiece.Coral);
+      scoringSubsystem.setTarget(intakeFieldTarget);
+      scoringSubsystem.fireTrigger(ScoringTrigger.BeginIntake);
+    }
+  }
+
+  public void execute() {
+    if (scoringSubsystem != null) {
+      // scoringSubsystem.setGamePiece(GamePiece.Coral);
+      // scoringSubsystem.setTarget(intakeFieldTarget);
```
jkleiber, line 46: can we remove these / comment why we need to fire the BeginIntake trigger continuously in execute?

Theoretically the scoring subsystem should be able to maintain its state without continual prompting so this seems sus

jkleiber (commented):
one comment

src/main/java/frc/robot/commands/strategies/AutoIntake.java
aidnem: I think the problem is, AutoIntake is entered on the same loop as scoring state machine realizes it doesn't have an object and goes back to idle. Because of this, the BeginIntake trigger gets lost because it hasn't yet gone from scoring back to idle. I could add a conditional transition from score to intake if we don't have an object to allow the frame-perfect transition instead, I'll try that first and then leave this in if it doesn't work.

aidnem: Epic gamer frame perfect intake is in and works

aidnem:
@jkleiber I made scoring less dumb and it still works, thanks for pointing that out

jkleiber (approved):
Approved

PR #83, opened 2025-02-13
Make elevator tunable and integrate elevator controller

aidnem:


aidnem:
mainly created this PR right now so I could see the diff and figure out what I broke while trying to integrate because right now CRT doesn't even work in sim

aidnem:
@jkleiber My goal for today is to verify that homing mode works and then merge this branch to avoid having another huge integration branch like we did in offseason.

jkleiber (approved):

PR #84, opened 2025-02-13
Add simplified motion profile for along track

linglejack06:
- scales velocity based on how far cross track we are (so we line up both directions in similar time)
- switch to translation2d get distance for comparing pose to desired pose


linglejack06:
@jkleiber these are changes from Monday shop.

Only thing still acting weird is the lineup activation after driving past reef. We can merge how it is and fix that in another pull request or fix it here? I just need someone to look over code to see why that might be happening @aidnem could also help check over if free

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -492,10 +492,14 @@ public boolean isDriveOTF() {
   public boolean isDriveCloseToFinalLineupPose() {
     // relative to transforms first pose into distance from desired pose
     // then get distance between poses (if less than 0.1 meters we are good)
+    Logger.recordOutput(
+        "Drive/distanceToLineupNewMethod",
+        this.getPose()
```
jkleiber, line 497: @linglejack06 have we tried logging getPose() to see if it keeps updating?

I think it would be acceptable to try to log getPose() and this distance in periodic() to help better understand why (and what) gets stale. Maybe it shows up in sim?

aidnem:
@linglejack06 can we still fix it by double tapping the trigger? I'll look over code and see if I can figure out why it happens.

jkleiber (approved):
I think this is good to merge even with the OTF issue, which is not related to this PR. I left a comment on a debug method but I think we will probably need to dig deeper

We should also check that getPose is being updated from the correct source of information (i.e. we need to be pulling that from telemetry but not the separate thread)

jkleiber:
@linglejack06 if you fix format I can reapprove

src/main/java/frc/robot/subsystems/drive/Drive.java
linglejack06: @jkleiber this logging (where i get the distance, constantly updates, and it is based on pose) therfore, i think we can assume get pose is updating. i feel the root issue lies somewhere in how the state machine is checking its conditionals

linglejack06:
> @linglejack06 can we still fix it by double tapping the trigger? I'll look over code and see if I can figure out why it happens.

yes that works perfectly, just slightly annoying to have to do that

Issue #76, opened 2025-02-10
Clean up drivetrain constants


jkleiber:
**Summary**
There are many constants in DrivetrainConstants.java/json related to velocity limiting that are unused or named poorly. The objective of this ticket is to go through and ensure only used constants are in these files, and that the constants have good names.

I also want the code to be more commented for each group of similar constants. This ticket also serves as a place to ensure the constants are realistic values. For example, these constants are not correct by inspection:
```
public final Double maxLinearSpeed = 6.0;
  public final Double maxAngularSpeed = Math.PI * 2;
  public final Double joystickDeadband = 0.1;
  public final Double maxPIDVelocity = Math.PI * 2;
  public final Double maxPIDAcceleration = Math.PI;
```

- for what it's worth joystickDeadband should probably be 0.17 to avoid the robot randomly slowly drifting while no control input is applied.
- maxPIDVelocity and maxPIDAcceleration are not descriptive variable names
- maxAngularSpeed and maxLinearSpeed are both likely wrong values (for sure on linear speed), but more importantly, it isn't obvious what mode of operation they apply to. We have a lot of max speed/accel constants that are for different modes of operation so we need to make these distinct.

**Work Scope**

- [x] Look through the DrivetrainConstants.java / json files and ensure unused constants are removed
- [x] As a second pass, check that all used constants have physically realizable values (i.e. max speed/accels are within robot's performance envelope)
- [x] Rename constants when needed to make it more clear what they are related to
- [x] Add short comments to similar blocks of constants to better organize the java file

PR #91, opened 2025-02-14
normalize drivetrain constants

linglejack06:
- add lineup constants to comp and hitl environments
- use json vision transofmr for real vision init


jkleiber (approved):
these are sane changes, so let's merge them

note: more fixes expected in #76

PR #94, opened 2025-02-15
63 add 3d robot to sim

avidraccoon:


jkleiber:
@avidraccoon when you get this working it would be helpful to update the repository README to explain how to set up the 3D sim

jkleiber:
@aidnem / @linglejack06 / @avidraccoon can we get this PR across the finish line today? I think we just need to resolve the merge conflicts

avidraccoon:
@jkleiber Where should I place the docs?

jkleiber:
@avidraccoon if you add them to the readme of this repository I think that would work fine

src/main/deploy/constants/test_drivebase/ClimbConstants.json
```diff
@@ -1,18 +1,18 @@
 {
   "restingAngle": {
     "value": 20.0,
-    "unit": "Degrees"
+    "unit": "Degree"
```
jkleiber, line 4: Is this really supposed to be "Degree"?

jkleiber (commented):
@aidnem could you take a look at this? At this point if everything runs I'm fine with it, but you and @linglejack06 will be the primary users of this at first. So imo y'all should be the ones to approve that it meets expectations

aidnem:
Sure I'll look over it.

src/main/deploy/constants/test_drivebase/ClimbConstants.json
aidnem: All of the pluralized units are just remaps of their non-plural counterparts, so I think sometimes when you autogenerate the JSON it writes it with the singular unit version. @avidraccoon can probably bring more clarity on this though.

avidraccoon: When I did units for json I always used the singular version

avidraccoon: WPILib Units have both, but json currently can handle writing both (if plural will write singular) but only read the singular. Because even for the plural units, when you get their name it is singular because they are created for the singular then just copied for the plural

advantagekit_configs/Robot_401_2025/config.json
```diff
@@ -0,0 +1,123 @@
+{
```
aidnem, line 1: @avidraccoon I'm not sure which of these is the Claw, but the zeroed rotation for it needs to be horizontal, 0.465719787 radians above the CAD model being "horizontal".

This can be seen from this image:
image
 
CTRE onboard closed-loop control for an arm requires that zero be when the arm is "horizontal," which in our case means that the center of mass is on the same height as the pivot point for the arm. That would look like this:

image

(red line is horizontal)

The claw should appear in that position when wrist angle is at 0.

The number 0.465719787 is arctan(2.326/4.268), since in the screenshot X is left/right and Z is up/down.

linglejack06:
@avidraccoon everything looks good on drive to me other than weird lineup behaviour

advantagekit_configs/Robot_401_2025/config.json
aidnem: This might mean setting the claw rotation to 180 - 26.683778230898 degrees (0.465... radians = 26.683778230898 degrees) = 153.316221769

By the way, for some reason SafeSearch hides the results of that subtraction, so maybe don't try to calculate it yourself on the chromebook???? No idea why this is the case, when I disable SafeSearch on my own machine literally the only result is a calculator but whatever, I won't dig deeper in case there is some legitimate reason that this subtraction is cursed.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -53,6 +63,41 @@ public class RobotContainer {
...
+    if (rampSubsystem != null) {
+      ramp_rotation = rampSubsystem.getPosition();
+    }
+    height = Math.min(height, 1.87);
+    double stage_one_height = Math.max(height - 0.55, 0.0);
+    double stage_two_height = Math.max(stage_one_height - 0.66, 0.0);
+    Logger.recordOutput(
+        "testingPose", new Pose3d(new Translation3d(0.0, 0.0, 0.0), new Rotation3d(0.0, 0.0, 0.0)));
+    Logger.recordOutput(
+        "componetPositions",
```
aidnem, line 87: spelling: component

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -53,6 +63,41 @@ public class RobotContainer {
...
+        new Pose3d[] {
+          new Pose3d(new Translation3d(0.05, 0.01, 0.9), new Rotation3d(0.0, ramp_rotation, 0.0)),
+          new Pose3d(
+              new Translation3d(-0.16, 0.31, 0.115), new Rotation3d(climb_rotation, 0.0, 0.0)),
+          new Pose3d(
+              new Translation3d(0.34, 0.12, height + 0.35),
+              new Rotation3d(0.0, -claw_rotation, 0.0)),
```
aidnem, line 94: Maybe the stuff mentioned in [this comment](https://github.com/team401/2025-Robot-Code/pull/94/files#r1963673715) should go here.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -91,10 +137,10 @@ public void configureAutos() {
...
     if (FeatureFlags.synced.getObject().runDrive) {
       drive = InitSubsystems.initDriveSubsystem();
-      if (ModeConstants.simMode == frc.robot.constants.ModeConstants.Mode.MAPLESIM) {
-        drive.setPose(
-            new Pose2d(Meters.of(14.350), Meters.of(4.0), new Rotation2d(Degrees.of(180))));
-      }
+      // if (ModeConstants.simMode == frc.robot.constants.ModeConstants.Mode.MAPLESIM) {
+      //   drive.setPose(
+      //       new Pose2d(Meters.of(14.350), Meters.of(4.0), new Rotation2d(Degrees.of(180))));
+      // }
```
aidnem, line 143: Why aren't we initializing drive to a pose anymore? I'm fine with this change if it makes sense but in that case we should remove this commented out code.

src/main/deploy/auto/ExampleAutoPath.json
```diff
@@ -1,6 +1,6 @@
 {
   "scoringLocations": [
-    "Reef2",
+    "Reef5",
```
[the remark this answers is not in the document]
avidraccoon: Was testing something and forgot to revert it

aidnem (changes requested):

aidnem:
Everything looks good except we need a correct claw angle, the auto shouldn't try to score the same location twice, and There was 1 spelling issue I found.

src/main/java/frc/robot/RobotContainer.java
jkleiber: We probably should uncomment this code since it's how the robot starts on the field in maple sim. If you don't start on the field you can never get on the field in maple sim

(It would be even better if this could start us on the alliance start line in maplesim based on a configurable starting location, but we can do that in another ticket)

avidraccoon: This code has no use because it does not set the position of maple sim just odometry position

aidnem (changes requested):
Everything looks good to me, once you add instructions to README.md I'll approve

aidnem:
@jkleiber If you don't want the instructions in README anymore we can go ahead and merge this, I verified it in sim by using it to find the setpoints for:
- L1
- L2
- L3
- L4
- L2 algae
- L3 algae
- Processor

jkleiber:
@aidnem I'm fine with doing instructions in a separate PR/repo now that they're in the discussion

aidnem (approved):

aidnem:
@avidraccoon thanks so much for setting this up, this is huge

PR #105, opened 2025-02-18
add dynamic coral algae stack obstacles

### LINKED ISSUE #101: Dynamically add keepout zones during auto
**Purpose**
I am concerned that we will inadvertently run into the coral/algae stacks that are pre-staged near the alliance station wall during auto. To avoid running into this stuff, we should dynamically add in keepout zones around those things while in autonomous

**Project Scope**
- [ ] Add a keepout zone to the alliance side coral/algae pre-staged stacks when in auto
- [ ] Show it working in simulation

linglejack06:
- add auto init and teleop init methods to drive to set dynamic obstacles

resolves #101


linglejack06:
this should work, need to add translation 2ds that represent the box we want around the stacks and then test in sim

linglejack06:
@avidraccoon will json sync work with the java Pair class

avidraccoon:
I think but I am not a hundred percent sure

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -316,6 +321,20 @@ public void configureStates() {
...
+  public void autonomousInit() {
+    localADStar.setDynamicObstacles(List.of(new Pair<Translation2d, Translation2d>(null, null)), getPose().getTranslation());
+  }
+
+   /**
+   * remove algae coral stack obstacles for on the fly
+   */
+  public void teleopInit() {
+    localADStar.setDynamicObstacles(List.of(new Pair<Translation2d, Translation2d>(null, null)), getPose().getTranslation());
+  }
```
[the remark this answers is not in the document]
linglejack06: @jkleiber rhey are for now, idea is we would find the squares we want removed for algae coral stack and add those in auto as dynamic obstacles, then teleop would remove them. The list of translation 2d pairs is a corner to corner square of where we don't want to go

I was confused about the get pose thing too though

aidnem:
@linglejack06 what's the status of this? Can we merge?

linglejack06:
@aidnem @jkleiber rhis just needs pairs of translation 2da that represent the upper left and bottom right corners of the boxes we want to avoid

jkleiber:
@linglejack06 can you make those boxes? There should be enough info based on sim and the field drawings

linglejack06:
Yeah I can get that done today

linglejack06:
@jkleiber @aidnem field coordinates added based off 2025 field drawing. y coordinates have a 0.3 meter span (so 15 cm on each side of center of coral stack (maybe we increase this). x coordinates  have a 0.2 meter span (so 10 cm on each side of center)

based off of this document: [https://firstfrc.blob.core.windows.net/frc2025/FieldAssets/2025FieldDrawings-FieldLayoutAndMarking.pdf](2025 Field Drawing)

jkleiber:
Yeah let's go up to 50 cm for now

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -319,6 +323,18 @@ public void configureStates() {
     stateMachine = new StateMachine<>(stateMachineConfiguration, DriveState.Joystick);
   }
 
+  /** add algae coral stack obstacles for on the fly */
+  public void autonomousInit() {
+    localADStar.setDynamicObstacles(
+        List.of(new Pair<Translation2d, Translation2d>(null, null)), getPose().getTranslation());
```
jkleiber, line 329: I think we need to actually set the obstacles still

avidraccoon:
Could we not make RedFieldLocations and BlueFieldLocations use the same
java class but initialized twice in JsonConstants once for each json file?

src/main/java/frc/robot/subsystems/drive/Drive.java
linglejack06: yep

jkleiber (approved):
approved

jkleiber:
> Could we not make RedFieldLocations and BlueFieldLocations use the same
java class but initialized twice in JsonConstants once for each json file?

@avidraccoon we could do that, and ideally we would. However, competition is in a few days, so that ship has sailed for now. Feel free to open a ticket and work the fix though!

PR #107, opened 2025-02-18
Transition to warmup if StartWarmup was fired during Init

aidnem:
Makes the transition for Seeded either go to warmup or idle depending on if StartWarmup was commanded during init.

Here's verification of it going straight from Init to Warmup during an auto in sim. I made it unable to seed by any method other than exceeding max time, and then made the max time like 9 seconds to ensure that warmup would be commanded before it finished seeding:

image


aidnem:
I need to verify that this doesn't break things if Init does actually finish in time, then this will be ready to merge.

aidnem:
Here's it going to idle like normal since it wasn't commanded to warmup:

image

linglejack06 (approved):
Lgtm

PR #110, opened 2025-02-20
Add TuningState to ScoringSubsystem

aidnem:
Replace "overrideStateMachine" with a TuningState. Tuning state does nothing (which means it prevents the setting of setpoints that normally occurs in idle). This allows the elevator to automatically home when test mode is enabled, and then go to Tuning mode. This is preferable to the old system, where enabling test mode would override state machine, preventing Init state from completing.


aidnem:
Here's enabling and disabling test mode a few times:

image

Last thing left to test is whether or not it can seed normally in teleop.

aidnem:
Here it is behaving normally in teleop:

image

aidnem:
This should be ready to merge, but I'd like to have someone double check my logic since there's a ton of boolean action going on in this PR.

src/main/deploy/constants/test_drivebase/ElevatorConstants.json
```diff
@@ -20,9 +20,9 @@
   },
   "elevatorCANCoderDiscontinuityPoint": 1.0,
   "elevatorLargeCANCoderID": 32,
-  "isLargeCANcoderClockwisePositive": true,
```
jkleiber, line 23: is this a correct translation? it seems like no, but it could be correcting for something we had wrong before

jkleiber (commented):
initial comments

src/main/java/frc/robot/subsystems/scoring/states/TuningState.java
```diff
@@ -0,0 +1,18 @@
...
+public class TuningState implements PeriodicStateInterface {
+  private ScoringSubsystem scoringSubsystem;
+
+  public TuningState(ScoringSubsystem scoringSubsystem) {
+    this.scoringSubsystem = scoringSubsystem;
+  }
+
+  @Override
+  public void periodic() {
+    if (!DriverStation.isTest()) {}
```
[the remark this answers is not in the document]
aidnem: Looked like it went back to Idle just fine because of checks in other places (ScoringSubsystem periodic), I removed this

src/main/deploy/constants/test_drivebase/ElevatorConstants.json
aidnem: This was incorrect before (sim had both encoders spinning in the same direction which makes no sense), I updated the constant format and also updated the values to be correct at the same time.

jkleiber (approved):

PR #112, opened 2025-02-21
100 add ramp intake test mode

avidraccoon:


aidnem:
image

Here's me changing the setpoint and then making kI 1 million and the sim responding, as verification the test mode works.

jkleiber (approved):
Thanks for knocking this out. Approved

PR #113, opened 2025-02-21
Climb Test Mode

minhnguyenbhs:
Hi!! @aidnem sorry could you take a look at this for me - the JSON Constants aren't working and I don't know why

other than that I think this should work?


src/main/java/frc/robot/subsystems/ramp/RampSubsystem.java
```diff
@@ -30,4 +30,9 @@ public void prepareForIntake() {
   public boolean isInPosition() {
     return mechanism.inPosition();
   }
+
+  public boolean isInClimbPosition() {
+    return mechanism.position == JsonConstants.rampConstants.climbPosition
```
jkleiber, line 35: This should already be handled by inPosition above. Check to see if it is.


(Side Note)
In general we can't compare doubles with ==. So to do this type of check we need to add a tolerance here:

```
Math.abs(mechanism.position - JsonConstants.rampConstants.climbPosition) < JsonConstants.rampConstants.climbPositionTolerance
```

This should be how inPosition does it already

src/main/java/frc/robot/subsystems/climb/ClimbSubsystem.java
```diff
@@ -21,7 +28,7 @@ public class ClimbSubsystem extends SubsystemBase {
   private ClimbInputsAutoLogged inputs = new ClimbInputsAutoLogged();
   private ClimbOutputsAutoLogged outputs = new ClimbOutputsAutoLogged();
 
-  // TODO: replace when ramp becomes a thing
+  private RampSubsystem rampSubsystem;
   private BooleanSupplier rampClear = () -> true;
```
jkleiber, line 32: Is rampClear being set? Why are we passing the whole rampSubsystem to climb rather than supplying isInPosition to this (or some climb specific variant if we don't care about being precisely in position)?

src/main/java/frc/robot/subsystems/climb/ClimbIOTalonFX.java
```diff
@@ -57,9 +57,9 @@ public ClimbIOTalonFX() {
                     .withKV(ClimbConstants.synced.getObject().climbkV)
                     .withKA(ClimbConstants.synced.getObject().climbkA)
                     .withKG(ClimbConstants.synced.getObject().climbkG)
-                    .withKP(ClimbConstants.synced.getObject().climbP)
-                    .withKI(ClimbConstants.synced.getObject().climbI)
-                    .withKD(ClimbConstants.synced.getObject().climbD))
+                    .withKP(ClimbConstants.synced.getObject().climbkP)
+                    .withKI(ClimbConstants.synced.getObject().climbkI)
+                    .withKD(ClimbConstants.synced.getObject().climbkD))
```
jkleiber, line 62: These JSON PID constants aren't working because they aren't named correctly in the JSON files. Also maybe this is a good time to switch the names over to the format: climbKp, climbKi, climbKd, etc.

src/main/java/frc/robot/subsystems/climb/ClimbSubsystem.java
```diff
@@ -82,6 +101,27 @@ public ClimbSubsystem(ClimbIO io) {
...
+        new LoggedTunableNumber("climbTunables/climbkD", ClimbConstants.synced.getObject().climbkD);
+
+    climbkS =
+        new LoggedTunableNumber("climbTunables/climbkS", ClimbConstants.synced.getObject().climbkS);
+    climbkV =
+        new LoggedTunableNumber("climbTunables/climbkV", ClimbConstants.synced.getObject().climbkV);
+    climbkA =
+        new LoggedTunableNumber("climbTunables/climbkA", ClimbConstants.synced.getObject().climbkA);
+    climbkG =
+        new LoggedTunableNumber("climbTunables/climbkG", ClimbConstants.synced.getObject().climbkG);
```
jkleiber, line 119: See my other comment about updating the JSON files to use the correct names for the pid gains

jkleiber (commented):
First pass review. Hopefully this helps!

aidnem:
Justin beat me to it on this one, I can help you look over later if stuff still doesn't work but it looks like he found the issue already.

minhnguyenbhs:
hi sorry if this was already finished at shop- i was a bit busier at jshs than i expected

image

minhnguyenbhs:
hello! the climb arm absolutely doesn't budge and i can't tell why. the cancoder works, the ids are correct in theory, and i did update the firmware- can you maybe tell me what is wrong? thank you!

src/main/java/frc/robot/constants/ClimbConstants.java
```diff
@@ -17,32 +19,35 @@ public class ClimbConstants {
...
-  public final Double climbI = 0.0;
-  public final Double climbD = 0.0;
+  public final Double climbkP = 5.0;
+  public final Double climbkI = 0.0;
+  public final Double climbkD = 0.0;
 
   public final Double climbCurrentLimit = 60.0;
 
-  public final Boolean invertFollowerClimbMotor = false;
+  public final Boolean invertFollowerClimbMotor = true;
```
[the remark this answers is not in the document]
linglejack06: fixed in latest commit

linglejack06:
@jkleiber this has been tested, just waiting on the test again today with updated setpoint then we can merge this as climb integrated

jkleiber (commented):
@linglejack06 did you guys get a chance to test this yet?

After testing, can you please go in and update the java constants as well? I also would like it if we feature flagged all the subsystems back on before this merges

aidnem:
We couldn't test it again because it needed to be reinforced and we didn't want to obliterate it again

jkleiber:
@aidnem @linglejack06 is this in a reviewable state / has it been tested with acceptable performance on the robot?

If the setpoints are kinda wrong but the controller works then we should merge this and find the constants later

aidnem:
Test mode didn't work but the climb itself would've worked if climbing was mechanically possible during test mode. @jkleiber I can either try to fix it tonight/tomorrow morning (although I'm also doing #119 which I think is higher priority) or we could go ahead and merge.

@PChild it depends on if you think climb will be working tomorrow to test

jkleiber:
@aidnem if it works in periodic let's ship it and open a deferred ticket for @minhnguyenbhs to fix test mode

I'm about to be on a plane to Roanoke rn so won't be able to approve right away. If you can't get an approval by the time I get home I'll take a look

PChild:
> Test mode didn't work but the climb itself would've worked if climbing was mechanically possible during testing with periodic. @jkleiber I can either try to fix it tonight/tomorrow morning (although I'm also doing #119 which I think is higher priority) or we could go ahead and merge.
> 
> @PChild it depends on if you think climb will be working tomorrow to test

We will have an updated climbing setup to test at some point tomorrow. The stow position and extend position change slightly and we will need to set a good climb position for it to pull to.

aidnem:
Looking at the test mode, I think it should actually work. I believe the problem was we weren't pressing the button to cance, allowing it to go back to idle and then transition to override mode for setpointing. In light of this I think we should keep this open until we get the corrected setpoints.

PChild:
Can we test it in sim

aidnem:
I got it to a point where I can move the goal setpoint in sim, the sim was astronomically strange (not sure if/how it ever worked), I'm trying to make sim actually able to control the climber arm now.

minhnguyenbhs:
sorry I thought that I'd verified that climb sim works, what is it doing right now?

aidnem:
@minhnguyenbhs it just doesn't move. It will apply up to 1.52 volts but the simulated climber never moves.

jkleiber:
@aidnem I heard this got tested yesterday and that the code worked. Can we get this merged?

PR #122, opened 2025-02-22
Prevent wrist slamming down on enable

aidnem:
Make scoring InitState set wrist goal angle to idle, instead of going to zero like it did before. This stops the wrist from slamming down during homing.

### DIFF
--- a/src/main/deploy/constants/config.json
+++ b/src/main/deploy/constants/config.json
@@ -1,5 +1,5 @@
 {
-  "environment": "comp",
+  "environment": "test_drivebase",
   "defaults": "comp",
   "environments": [
     {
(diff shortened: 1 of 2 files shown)


aidnem:
I can verify this works in sim (behavior makes sense in 3D sim and the wrist position remains steady), I won't screen record because recording and compressing video is a huge pain but here's the graphs showing that the goal angle remains correct during homing and into idle:

image

The reason the position is wrong at the start is because when the robot is disabled the wrist falls all the way down in sim because it doesn't know about the hardstop situation.

src/main/deploy/constants/config.json
```diff
@@ -1,5 +1,5 @@
 {
-  "environment": "comp",
+  "environment": "test_drivebase",
```
jkleiber, line 2: in the future can we stay on comp?

If comp doesn't work in the sim, then that's something we should fix

jkleiber (approved):
Approved. Please see my 1 comment for future reference

src/main/deploy/constants/config.json
aidnem: None of the comp gains work in sim.

PR #124, opened 2025-02-22
Add reef avoidance to the wrist

aidnem:
If drive is within a certain distance of the wrist and the elevator is trying to move past the reef, clamp wrist to be in a position where it won't hit the reef, and if wrist is in an unsafe position, stop elevator from moving past a reef level until it is in a safe position.


aidnem:
We can test this at shop and then merge once we've verified that it works, or I could test in 3D sim if you trust it enough. The values need to be tuned though.

jkleiber (commented):
looks reasonable to test. Can we remove all the TODO comments for gain tuning?

aidnem:
@jkleiber I removed to-dos that were done, I think we just need to find the reef base avoidance height (which only comes into play in test mode anyway) and it be done. Is there anything else you want before we merge?

jkleiber:
@aidnem that sounds good to me

aidnem:
Tonight at shop we determined some more clamps (unfortunately not reef height because we broke the cable chain slightly while going for net) and also found the setpoints for L2, L3, L4, L2 algae, L3 algae, processor, and net using the new claw. I put those setpoints in this branch since this branch needed us to find certain setpoints anyway.

I have a reef height value that I made up by subtracting a little bit from the lowest value we'd NEED to move the wrist down at (L2 algae), so this value should work. We could always test this out later, or we could not merge until we're 100% sure.

@jkleiber would you mind looking this over so it can get merged soon?

src/main/deploy/constants/ScoringSetpoints.json
```diff
@@ -101,22 +101,22 @@
...
-      "value": -0.75,
+      "value": -0.15,
       "unit": "Rotation"
     }
   },
   "net": {
     "name": "net",
     "elevatorHeight": {
-      "value": 1.25,
+      "value": 1.9,
```
jkleiber, line 115: Do we need full extension to score in the net? Just worried about hitting the hard stops

src/main/deploy/constants/comp/WristConstants.json
```diff
@@ -49,9 +49,17 @@
     "unit": "Radian"
   },
   "minElevatorDownSafeAngle": {
-    "value": 0.0,
+    "value": -0.503,
```
jkleiber, line 52: This seems sus

jkleiber (commented):
First pass. Apparently I was mid review a few days ago and got distracted and forgot to submit the comments

src/main/deploy/constants/comp/WristConstants.json
aidnem: Oops I think I put a radian value as being in rotations

src/main/deploy/constants/ScoringSetpoints.json
aidnem: Yes, this position seemed to be as low as we could go (maybe we could go a tiny bit lower but it would be sketchy to roll the algae out of the claw at a higher angle). It was about an inch away from the hardstop when we tested this last night.

src/main/deploy/constants/comp/ElevatorConstants.json
```diff
@@ -56,30 +56,50 @@
     "value": -3.0,
     "unit": "Volt"
   },
-  "homingVelocityThresholdMetersPerSecond": 0.1,
+  "homingVelocityThresholdMetersPerSecond": 0.001,
```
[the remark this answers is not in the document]
aidnem: I was unaware I changed this, must've accidentally written it using the java file and not noticed it was wrong

src/main/deploy/constants/comp/ElevatorConstants.json
```diff
@@ -62,24 +62,28 @@
     "unit": "Second"
   },
   "homingMaxUnmovingTime": {
-    "value": 0.25,
+    "value": 1.0,
```
jkleiber, line 65: Should this go back to 0.25?

src/main/java/frc/robot/constants/WristConstants.java
```diff
@@ -53,42 +54,44 @@ public class WristConstants {
...
+  public final Double wristKG = 0.34;
+  public final Double wristKS = 0.2;
+  public final Double wristKV = 1.0;
   public final Double wristKA = 0.0;
 
-  public final Double wristKP = 0.0; // TODO: Tune these
+  public final Double wristKP = 20.0;
   public final Double wristKI = 0.0;
-  public final Double wristKD = 0.0;
+  public final Double wristKD = 0.5;
```
jkleiber, line 64: Can we update these to match the JSON now?

jkleiber (commented):
Second pass

aidnem:
@jkleiber constants should be good now 👍

PR #127, opened 2025-02-23
add final snakescreen tables to robot

linglejack06:
- add game piece selector
- add publishers for each snakescreen value (so snakescreen recognizes robot defaults)


linglejack06:
@PChild ive been able to publish all fieldx except for the reef location, which refuses to update. I think this might be because javascript useEntry wants a number type, but i can only send either Double or Integer, do you have any ideas?

linglejack06:
As far as I know the network tables value is set, it's just not getting updated on snakescreen ui

linglejack06:
> As far as I know the network tables value is set, it's just not getting updated on snakescreen ui 

This makes sense as react will only update if there is a change to the mapped entries, but the network tables here touch the scoring target value, I think I can add this change to a branch in snakescreen to fix it

linglejack06:
> Huh, that should just work. We can look at it in shop tonight. Worst case we just switch it to string on both sides so there's no compatibility issue?

Yeah we could do that to hopefully fix it. My only concern is that react isn't seeing the changes, because maybe the way the list is set up to read the entry's and barely depend on the state makes it so where it isn't refreshing when it has that change in state from network tables.

I'll play with it if I have time before shop and get your help at shop tonight

avidraccoon:
Did you check advantage scope?

linglejack06:
Yeah the values are getting published. Snakescreen just isn't seeing them or isn't reloading to reflect them.

avidraccoon:
Do the other values update on snake screen?

avidraccoon:
It is supposed to update at about 60 times per second by default

linglejack06:
Yes autonomy mode has been demonstrated to work, haven't tested coral / algae selection. I think I'll switch to using strings for snakescreen and robot code, maybe that fixes things.

I bet it's just a weird bug going on with react state management

avidraccoon:
It might just be js being stupid with how it handles numbers

avidraccoon:
If so I should not be that hard to fix

linglejack06:
Yes I'm thinking it's that OR react just isn't realizing it needs to refresh the list of buttons, since the scoring target is only mentioned once in there. And react only reloads components that mention the state that changed. Both are simple fixes though, I'll just have to see what works tonight at shop or whenever I have the time since we probably will be tuning lineup and score tonight

avidraccoon:
Might be because they dont have keys and react is not happy when list of elements dont have keys

linglejack06:
That’s very true Sent from my iPhoneOn Feb 24, 2025, at 2:45 PM, avidraccoon ***@***.***> wrote:﻿
  Might be because they dont have keys and react is not happy when list of elements dont have keys—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>



avidraccoon left a comment (team401/2025-Robot-Code#127)
Might be because they dont have keys and react is not happy when list of elements dont have keys

linglejack06:
Update: snakescreen reef now publishes in sim. I just want to fully test this on robot before we merge these changes. @jkleiber do you want reef locations to update during auto (so you could visualize where it's going next on the screen)

avidraccoon:
@linglejack06 Double work?

jkleiber:
@linglejack06 sure let's have it publish in auto as well

linglejack06:
updated

PR #129, opened 2025-02-23
Adds LEDs

sleepyghost-zzz:
Added Logic for LEDs


AdvantageKit-License.md
jkleiber, line 1: Please remove this file

build.gradle
```diff
@@ -29,16 +29,16 @@ deploy {
...
+                    // // https://www.chiefdelphi.com/t/2024-wpilib-feedback/464322/141
+                    // final MAX_JAVA_HEAP_SIZE_MB = 100;
+                    // jvmArgs.add("-XX:+UnlockExperimentalVMOptions")
+
+                    // // Set the minimum heap size to the maximum heap size to avoid resizing
+                    // jvmArgs.add("-Xmx" + MAX_JAVA_HEAP_SIZE_MB + "M")
+                    // jvmArgs.add("-Xms" + MAX_JAVA_HEAP_SIZE_MB + "M")
+                    // jvmArgs.add("-XX:GCTimeRatio=5")
+                    // jvmArgs.add("-XX:+UseSerialGC")
+                    // jvmArgs.add("-XX:MaxGCPauseMillis=50")
```
jkleiber, line 41: Please undo this change or provide a reason for why it was necessary (I don't have context on why this would have changed)

settings.gradle
```diff
@@ -4,7 +4,7 @@ pluginManagement {
     repositories {
         mavenLocal()
         gradlePluginPortal()
-        String frcYear = '2025'
+        String frcYear = '2024'
```
jkleiber, line 7: This should be 2025

src/main/java/frc/robot/Constants.java
jkleiber, line 1: Please remove this flie

src/main/java/frc/robot/constants/FeatureFlags.java
```diff
@@ -19,7 +19,9 @@ public final class FeatureFlags {
   public final Boolean runDrive = true;
   public Boolean runVision = true;
   public final Boolean runScoring = false;
+  public boolean runLEDs = true;
```
jkleiber, line 22: This needs to be capital B Boolean for JSON to work

src/main/java/frc/robot/generated/TunerConstants.java
jkleiber, line 1: Please remove this file

src/main/java/frc/robot/subsystems/drive/ModuleIOSimMaple.java
jkleiber, line 1: For consistency with other IOs, can we call this file/class "ModuleIOMapleSim"

src/main/java/frc/robot/subsystems/led/LED.java
```diff
@@ -0,0 +1,221 @@
...
+      if (scoringSubsystem != null && scoringSubsystem.getTarget() == FieldTarget.L3) {
+        addSplitPattern(targetOnReefL3, clear);
+      }
+      if (scoringSubsystem != null && scoringSubsystem.getTarget() == FieldTarget.L4) {
+        addSplitPattern(targetOnReefL4, clear);
+      }
+      if (climbSubsystem != null && climbSubsystem.getLockedToCage()) {
+        addPattern(lockedOnHang);
+      }
+      if (driveSubsystem != null && driveSubsystem.isDesiredLocationReef()) {
```
jkleiber, line 102: This logic is a bit confusing to read for me. Also, by chaining if statements together you are implicitly creating a priority order of what LED pattern to show. 

In this case, regardless of what is happening with coral or algae or targeted reef level, if the robot is going to the reef the L1 target LEDs will be what light up. This is misleading in many ways. I think it would probably be better to create a markdown doc where you explain what LED patterns are available, when each get shown, and which has priority over others. Then it will be easier to implement that logic in code with if/else-if/else cases

vendordeps/NavX.json
jkleiber, line 1: Please remove this file. We don't have a NavX

vendordeps/Pheonix6-release.json
jkleiber, line 1: I think this is an outdated version of this vendor dependency. Please remove it

jkleiber (commented):
First pass comments.

There were a lot of random files in this PR that don't make sense, some of which are things that ship with a default project. In the future, please take a look at the PR files changed to see if there are weird files appearing

Also, I hope my comment in the LED subsystem about organizing your thoughts / designing the logic more is helpful. Please lmk if you have any questions!

src/main/java/frc/robot/subsystems/led/LED.java
sleepyghost-zzz: I understand the markdown doc completely, but with using multiple patterns that apple to a list over ``periodic()`` how would i make it not be a priority list, or should i just make it a actually priority list instead of a jumble of if statements

jkleiber: What I want to know is what the priority list is. I also think it makes more sense to show as much info as possible 

For example, the following simple scheme seems reasonable to me:

Top third of the strip is pink if we have coral, green if we have algae, white if both, and nothing if neither

The bottom two thirds of the strip shows our reef level with 4 distinct colors / color sizes

The whole strip flashes red once at 20 seconds remaining 

The bottom third of the strip shows blue if the beelink is disconnected at any point 

During disabled we run orange lasers if the robot is in coast mode instead of brake mode


You'd basically want to keep track of the thirds and then set them in the corret order in the if statements. This way the thirds of the strip always mean the same thing and don't randomly get overwritten by other information.

You don't have to use the scheme above, but I and the drivers would appreciate if the codes were super easy to remember

jkleiber (commented):
Thanks for adding the markdown. What I'm really wanting to know is a mapping of codes to meaning. I also don't want reef level to overwrite game piece or vice versa. We should be able to see the relevant robot state immediately without having to worry about overriden codes

There will be some codes that take priority (error detected, 20 seconds left) but these should be temporary or need to be constantly alerting

avidraccoon:
@jkleiber  Do you think we should add addRightPattern and addLeftPattern functions?

sleepyghost-zzz:
@avidraccoon @jkleiber nathan right now there is a function called runSplitPattern that takes in two patterns and puts them on left and right data, so you would just need  copy paste that. if you want

minhnguyenbhs:
@sleepyghost-zzz hi! coach justin said that the leds were acting oddly yesterday and the colors were split up incorrectly; was one of the thirds just in the wrong place? should i be simulating it with 180 columns or 2 columns (it seems to work fine in 180 but not with 2?)

minhnguyenbhs:
also what are serpentine leds?

minhnguyenbhs:
the problem might just be that sim had serpentine turned on and it shouldn't have
i'm unsure if this actually is the issue but it would be nice to test in shop next time :D

jkleiber (approved):

PR #140, opened 2025-02-28
Add competition autos

aidnem:
Adds:
 - 1 Coral Left side
 - 1 Coral barge side
 - 2 Coral left side
 - 2 Coral right side
 - 3 Coral left side
 - 3 Coral barge side (fills barge face, then puts one in barge/right side face)
 - 4 Coral left side


aidnem:
These need to be tested in sim before this can get merged

aidnem:
@jkleiber I changed my mind on how I'd prefer to order these changes. I got this auto to work in sim (it took forever due to poorly tuned lineup, but the scoring conditions were eventually met, it went to the next phase, etc.) so I think the logic on this branch is passable. Could we merge this today and then bring these changes over to lineup-fix? Since we know that this works (a little suspiciously) in sim, we could then know that the only thing holding it back from working in real life is controller issues/drive state machine issues and continue lineup fixing on the lineup fix branch.

Let me know if you disagree with this, I'd be glad to do something else.

jkleiber (approved):
Approved, with the note that some of these autos will need to change because they are likely going to wild scoring locations from a speed perspective

PR #149, opened 2025-03-04
Improve algae reliability

aidnem:
- Separate algae and coral intake voltages
- Increase algae score voltage to 12
- When scoring in the net:
  - Warmup takes elevator to a height around L2 height and waits
  - Score commands elevator all the way up, as soon as elevator is near the top applies 12V to kick the algae out of the claw using vertical momentum from elevator

This PR should also be used to fix any algae reliability issues (state machine was getting stuck but I think that's fixed) so will remain as a bit of an integration branch (should be pretty quick to verify).

### DIFF
--- a/src/main/deploy/constants/ScoringSetpoints.json
+++ b/src/main/deploy/constants/ScoringSetpoints.json
@@ -120,6 +120,17 @@
       "unit": "Rotation"
     }
   },
+  "netWarmup": {
+    "name": "netWarmup",
+    "elevatorHeight": {
+      "value": 0.25,
+      "unit": "Meter"
+    },
+    "wristAngle": {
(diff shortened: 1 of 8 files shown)


aidnem:
This is waiting on #143 because this branch is based off of that branch, I'm not sure what would happen if this were merged first.

jkleiber:
I'm pretty much good with this change, will approve after we see it in shop + get the other branch merged

aidnem:
I unfortunately cannot come to shop tonight but the setpoint I put should actually be physically possible so if you test this it should ideally go smoothly

jkleiber (approved):
approved. Let's actually merge this to make integration easier when this gets tested at shop

PR #150, opened 2025-03-04
Add far warmup to scoring

aidnem:
Adds a new state to scoring in between idle and warmup that can be used to raise the elevator a little to shorten warmup times for L3 and L4 scoring.


aidnem:
image

Here's elevator rising before OTF finishes in auto in sim

jkleiber (approved):
Approved

PR #151, opened 2025-03-04
Auto integration

linglejack06:
- ensure climb works (?)
- make ramp work with climb (?)
- check that algae works (?)
- get three piece auto (?)
- algae works with mixed autonomy (?)


jkleiber:
@linglejack06 can you push your code

linglejack06:
@jkleiber i just realized I forgot to push from hotel laptop. It won't prevent me from doing anything in sim though, so I'll push it once I get to shop tomorrow night

linglejack06:
Thank you

linglejack06:
Also, we need to figure out why we were exiting lineup sometimes, were we really not getting an estimate for ten loops?

jkleiber:
@linglejack06 instead of exiting lineup due to a lack of observations, could we just stay in lineup until:
-  canceled by the driver (by giving up on the score) 
- if the robot gets disabled 
- if the autonomy mode becomes manual
- if a new OTF pose is requested and it's far away

Then in auto we would stay in lineup even if we lose measurements for a bit longer than 10 cycles

jkleiber:
We also need to make sure we are actually receiving observations. It's possible they are being filtered out erroneously, which would be a big problem in vision

That said, if we see this issue in sim we can rule that out probably

linglejack06:
@jkleiber really weird, but lineup is still cancelling after about 0.5 seconds even though ive set max observation age to 100 (so it shouldnt cancel at all)

also newest updates to otf seem to be working so far in sim, but even before updates i was never running into errors in sim so ill keep trying

linglejack06:
> @jkleiber really weird, but lineup is still cancelling after about 0.5 seconds even though ive set max observation age to 100 (so it shouldnt cancel at all)
> 
> also newest updates to otf seem to be working so far in sim, but even before updates i was never running into errors in sim so ill keep trying

acutally i updated max observation age in wrong constants folder. with new observation age it just constantly swings around tag never settling (probably because its only getting one observation

linglejack06:
otf is failing after the first try every single time. i think its auto exiting because drive to pose is returning null

linglejack06:
updated logic seems to work (not rescheduling ever, but also removed the drive is close to final lineup pose check above (maybe that was cancelling otf) lineup still suspicious but hard to test in sim with crappy vision sim (it just oscilates due to observations always being different)

linglejack06:
> We also need to make sure we are actually receiving observations. It's possible they are being filtered out erroneously, which would be a big problem in vision
> 
> That said, if we see this issue in sim we can rule that out probably

observations are being seen essentially every loop, no clue why its exiting early, but doesnt appear to be happening in sim unless we actually lose an estimate. However, its definitely now waiting 100 tries before cancelling. i think something is up with out conditional check there

jkleiber:
@linglejack06 I'm testing some in sim locally now and will take a look at lineup. That said you should take a look when you get a chance because my availability is spotty

linglejack06:
> @linglejack06 I'm testing some in sim locally now and will take a look at lineup. That said you should take a look when you get a chance because my availability is spotty

ok thank you, ill work more soon did your commit help anything?

linglejack06:
yeah that definitely is happening, once you get that commit in ill pull and
continue working on debugging. I think im getting close with otf debugging
might just need more conditionals for edge cases

linglejack06:
> @linglejack06 just pushed an update to make maplesim sort of match the comp bot. Now we are able to drive with comp config, but the lineup is still bad
> 
> The controller is oscillating but I'm not sure that's the issue. The observation is only valid for a very short period of time. So I think our observation age logic is correct.
> 
> On the OTF side I haven't seen us fail to go to an OTF pose yet, so that's encouraging. In my other commit I added a step to clear the queue in case there are any stale actions at autonomousInit and nulled out the command/action like you did for teleop and it's been good since then.

yeah i think thats helping otf, i think lineup is getting screwed in sim due to controller, and its never finishing due to vy not being low enough. also the one time i got lineup to be fully true, scoring was stuck in warmup

jkleiber:
If increasing the weight doesn't work, you can bring in the camera switch thing

linglejack06:
@jkleiber what if we just use regular instead of maple sim?

linglejack06:
@jkleiber adding second camera prevents us from leaving lineup and going way far away. but we still oscillate, especialyl whenw e switch to other camera robot moves like half a meter cross track away, then gains back the proper camera and lines up again

linglejack06:
ok, lineup taking forever but does eventually finish, it is just getting stuck there because scoring isnt going into score state

jkleiber:
@linglejack06 have you tried adding weight to the robot? it is trivially easy - just go to InitSubsystems.java and make the robot like 200 pounds. Add more weight until the oscillations go away

I'd like to not introduce the camera switch thing until we think our nominal logic is working

linglejack06:
latest commit changed observation age down to 10, after 10 we use the other camera observation (this actually works well in sim, but unsure if it will still lead to the crazy oscilations in real life)

ill check the weight thing and remove camera on my end

jkleiber:
Okay cool, so what do we need to do to make the robot score in sim (why isn't it going to scoring state?)

jkleiber:
@linglejack06 @aidnem FYI: I'm never going to buy a "comp gains don't work in sim" argument. If they aren't working in sim, we need to change the physics model in sim to be more accurate to the real robot

linglejack06:
> @linglejack06 @aidnem FYI: I'm never going to buy a "comp gains don't work in sim" argument. If they aren't working in sim, we need to change the physics model in sim to be more accurate to the real robot

so maybe just needs better elevator model, ill check that out. and ill log the scoring stuff to see what fails

linglejack06:
> @linglejack06 have you tried adding weight to the robot? it is trivially easy - just go to InitSubsystems.java and make the robot like 200 pounds. Add more weight until the oscillations go away
> 
> I'd like to not introduce the camera switch thing until we think our nominal logic is working

increased weight still had oscillation off to one side resulting in camera loss, i think this is just from relying on the old observation for too long, but we wouldnt have to if we had valid observations

jkleiber:
@linglejack06 
Let's not spend effort on the other camera thing until we understand why the single camera observations are invalid

I recommend setting up a hack that allows you to log the observations even when in Joystick mode for whatever reef face is selected (you could hardcode this to be convenient). Then in teleop, move the robot to where we think it should be able to see a tag and have it sit still. See if we randomly lose the tag / why we can/can't see the tag. Drive the robot around to see if we can lose that tag in sim and add logging to see why we lose it

jkleiber:
@linglejack06 another temporary solution to try: make the lineup crosstrack gains way lower and see if it works in sim.

aidnem:
The only thing required to make it work properly in sim should be inverting the direction of the large encoder's movement in ElevatorIOSim. Negating the position of the large encoder sim state's position should make the elevator actually travel the right direction (currently I think it goes through the bottom of the robot).

linglejack06:
> @linglejack06 Let's not spend effort on the other camera thing until we understand why the single camera observations are invalid
> 
> I recommend setting up a hack that allows you to log the observations even when in Joystick mode for whatever reef face is selected (you could hardcode this to be convenient). Then in teleop, move the robot to where we think it should be able to see a tag and have it sit still. See if we randomly lose the tag / why we can/can't see the tag. Drive the robot around to see if we can lose that tag in sim and add logging to see why we lose it

yeah thats what i was thinking too, il work more on this after school

aidnem:
@linglejack06 @jkleiber what if instead of holding old observations for a comically long time, we:

- Cache old observations for a short period of time (e.g. 10 periodic loops) and control based on the last observation (maybe even adjusted for odometry if we wanted to get really fancy, see below)
- After the old observation expires, assume 0, 0 and stop moving (this would fix the issue where it loses observation and drives in one direction since that's what the last observation told it to do)
- Wait for a valid observation with error below thresholds to determine that lineup was finished

I also think adjusting for odometry could be valuable when using old observations, though I'm not entirely sure how it could work. I'll read drivetrain code and see if this would be possible/easy enough to write and test in the 1 afternoon that we have). The general idea is:
  - whenever we get an observation, reset our "offset" to zero.
  - Whenever we don't have an observation, keep track of our odometry "offset" and try to make it match the negative error from vision (e.g. if our last measurement had a crosstrack offset of 0.5 meters, try to be at -0.5 meters from where we were when that observation was had).

linglejack06:
> @linglejack06 @jkleiber what if instead of holding old observations for a comically long time, we:
> 
> 
> 
> - Cache old observations for a short period of time (e.g. 10 periodic loops) and control based on the last observation (maybe even adjusted for odometry if we wanted to get really fancy, see below)
> 
> - After the old observation expires, assume 0, 0 and stop moving (this would fix the issue where it loses observation and drives in one direction since that's what the last observation told it to do)
> 
> - Wait for a valid observation with error below thresholds to determine that lineup was finished
> 
> 
> 
> I also think adjusting for odometry could be valuable when using old observations, though I'm not entirely sure how it could work. I'll read drivetrain code and see if this would be possible/easy enough to write and test in the 1 afternoon that we have). The general idea is:
> 
>   - whenever we get an observation, reset our "offset" to zero.
> 
>   - Whenever we don't have an observation, keep track of our odometry "offset" and try to make it match the negative error from vision (e.g. if our last measurement had a crosstrack offset of 0.5 meters, try to be at -0.5 meters from where we were when that observation was had). 

This is a good idea and I can try it out. I just feel like it's weird that we are ever losing the estimate anyways, photon vision in real life was seeing the tag. Could it just be that beelink isn't sending every loop?

I'm open to trying this idea as a work around though as we don't have time to fix every problem to a perfect degree

jkleiber:
@linglejack06 the cameras don't update at 50Hz and the beelink doesn't likely send packets at 50Hz over the network

@aidnem I think the lineup offset thing is a good idea

linglejack06:
@aidnem feel free to implement if you have time, if not I can implement around 3 after school

aidnem:
We don't even have to manually use odometry: we could keep track of global pose estimate as our "start position" whenever we get a valid estimate. Then on each frame we can take our global position, subtract our start position, and then rotate it by robot rotation to get it back into robot relative coordinates (I think?). This should be easier to implement too.

aidnem:
I'll try to implement this right now and then test in sim

jkleiber:
@aidnem I think the math is pretty much what you described. Let us know how it goes - I have meetings for a lot of the afternoon so I won't be able to run sims

linglejack06:
@aidnem im back at computer now lmk if you need help or need me to finish if you get busy

jkleiber:
@aidnem I like what you have written, and I recommend looking into taking it a step further

Since we are in lineup for such a short amount of time, what if we switch immediately over to odometry based on the adjustment thing if we don't have measurements rather than caching the measurement? Then we should set it to never cancel lineup if we've had a valid tag observation at any point in lineup, i.e.

```
if (observation.isValid()) // use the observation and set it as latest

else if (!observation.isValid()) // use the lastObservation and our pose at that time as an anchor (what you've done already)

else // We have no anchor and no valid measurement. We should probably go back into OTF here because we're totally in the wrong spot, but for now we can continue to cancel
```

jkleiber:
also it looks like sim wrist is failing because it has wrong maximum angle (or because we are commanding it the wrong way in sim for some reason)

linglejack06:
> @aidnem I like what you have written, and I recommend looking into taking it a step further
> 
> Since we are in lineup for such a short amount of time, what if we switch immediately over to odometry based on the adjustment thing if we don't have measurements rather than caching the measurement? Then we should set it to never cancel lineup if we've had a valid tag observation at any point in lineup, i.e.
> 
> ```
> if (observation.isValid()) // use the observation and set it as latest
> 
> else if (!observation.isValid()) // use the lastObservation and our pose at that time as an anchor (what you've done already)
> 
> else // We have no anchor and no valid measurement. We should probably go back into OTF here because we're totally in the wrong spot, but for now we can continue to cancel
> ```

how is this different from what aiden did @jkleiber ill add it real quick i just cant tell what you mean

jkleiber:
@linglejack06 because with Aiden's code we will still cancel lineup if we don't see an observation in 10 cycles. What I'm recommending is to continue with lineup as long as you have a latestObservation because you now know where you were and can rely on odometry

Basically don't cancel lineup unless you have no choice

linglejack06:
this also simplifies conditionals a lot, so should be easier to read and debug if we still have problems

src/main/java/frc/robot/subsystems/drive/states/LineupState.java
```diff
@@ -332,6 +343,36 @@ public DistanceToTag tryOtherCamera(
...
+    Translation2d adjustment =
+        drive
+            .getPose()
+            .minus(poseAtLastObservation)
+            .getTranslation()
+            .rotateBy(getRotationForReefSide().times(-1.0));
+
+    Logger.recordOutput("Drive/Lineup/Adjustment", adjustment);
+    Logger.recordOutput(
+        "Drive/Lineup/adjustedPose", drive.getPose().getTranslation().plus(adjustment));
```
aidnem, line 366: Just realized this line is really wrong and will give us garbage data (field relative global pose added to a robot relative translation).

I'll add a visualization of lineup's target instead

aidnem:
I was having a slight issue where robot would slam into reef in sim, not sure what the deal is there. Also, I think we need to move our camera offsets forward in sim because alongtrack is never satisfied that it's at the correct value

aidnem:
@linglejack06 We should test in shop tonight whether or not cached position is reliable enough to determine if lineup is finished, too. I feel like if our math is correct and the single pose estimates aren't too terrible when they have high error, it should be decently reliable.

jkleiber (approved):

PR #158, opened 2025-03-06
Elevator setpoint fix

linglejack06:
- add new elevator setpoints
- add ALgae desired locations to make intaking algae at reef actually work

### DIFF
--- a/src/main/java/frc/robot/InitBindings.java
+++ b/src/main/java/frc/robot/InitBindings.java
@@ -12,7 +12,6 @@ import frc.robot.constants.OperatorConstants;
 import frc.robot.subsystems.climb.ClimbSubsystem;
 import frc.robot.subsystems.climb.ClimbSubsystem.ClimbAction;
 import frc.robot.subsystems.drive.Drive;
-import frc.robot.subsystems.drive.Drive.DesiredLocation;
 import frc.robot.subsystems.drive.Drive.DriveTrigger;
 import frc.robot.subsystems.ramp.RampSubsystem;
 import frc.robot.subsystems.ramp.states.RampState.RampTriggers;
@@ -58,6 +57,7 @@ public final class InitBindings {
                       // And then fall through to the mixed autonomy behavior (no break here is
                       // intentional)
                     case Mixed:
+                      drive.setGoToIntake(false);
                       drive.fireTrigger(DriveTrigger.BeginOTF);
                       break;
                     case Manual:
@@ -157,7 +157,6 @@ public final class InitBindings {
                       // intentional)
                     case Mixed:
                       // Start auto align if in mixed autonomy
-                      drive.setDesiredIntakeLocation(DesiredLocation.CoralStationRight);
                       drive.setGoToIntake(true);
                       drive.fireTrigger(DriveTrigger.BeginOTF);
                       // Then always start intake for scoring (no break here is intentional)
--- a/src/main/java/frc/robot/subsystems/drive/Drive.java
+++ b/src/main/java/frc/robot/subsystems/drive/Drive.java
@@ -49,7 +49,6 @@ import frc.robot.subsystems.drive.states.IdleState;
 import frc.robot.subsystems.drive.states.JoystickDrive;
 import frc.robot.subsystems.drive.states.LineupState;
 import frc.robot.subsystems.drive.states.OTFState;
-import frc.robot.subsystems.scoring.ScoringSubsystem;
 import frc.robot.util.LocalADStarAK;
 import java.util.List;
 import java.util.concurrent.locks.Lock;
@@ -363,11 +362,11 @@ public class Drive implements DriveTemplate {
   @Override
   public void periodic() {
     // Manually cancel go to intake if we have a gamepiece
-    if (goToIntake && ScoringSubsystem.getInstance().isCoralDetected()) {
-      setGoToIntake(false);
-    } else if (goToIntake && ScoringSubsystem.getInstance().isAlgaeDetected()) {
-      setGoToIntake(false);
-    }
+    // if (goToIntake && ScoringSubsystem.getInstance().isCoralDetected()) {
+    //   setGoToIntake(false);
+    // } else if (goToIntake && ScoringSubsystem.getInstance().isAlgaeDetected()) {
+    //   setGoToIntake(false);
+    // }
 
     odometryLock.lock(); // Prevents odometry updates while reading data
     gyroIO.updateInputs(gyroInputs);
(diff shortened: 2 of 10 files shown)


linglejack06:
@aidnem @jkleiber could someone check this over to make sure ive added the new algae locations properly everywhere, this will make algae intake work without a weird hack

also net-otf branch just needs net and processor poses if we want to grab those tonight (with time) but id like to merge this first then rebase net-otf off main

linglejack06:
nothing appears to be crashing in sim, so i assume i added all algae locations correctly, but i dont have keyboard inputs to check algae intake stuff (unless you know what button corresponds to right trigger @aidnem ?)

src/main/java/frc/robot/InitBindings.java
```diff
@@ -157,7 +157,6 @@ public static void initDriveBindings(Drive drive, StrategyManager strategyManage
                       // intentional)
                     case Mixed:
                       // Start auto align if in mixed autonomy
-                      drive.setDesiredIntakeLocation(DesiredLocation.CoralStationRight);
                       drive.setGoToIntake(true);
```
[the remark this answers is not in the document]
linglejack06: what changes do i need to make to drive to enable this? does scoring care about drive lined up when algae intaking, or its locaiton at all?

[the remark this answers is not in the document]
linglejack06: looks like scoring will just intake until it has piece, so we could run this when we enter lineup for drive or it probably wouldnt even matter if we just fire intake trigger for scoring when we start otf (its reduced acceleration and elevator wont be that high)

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -364,8 +363,11 @@ public void teleopInit() {
   @Override
   public void periodic() {
     // Manually cancel go to intake if we have a gamepiece
-    if (goToIntake && ScoringSubsystem.getInstance().isCoralDetected()
-        || ScoringSubsystem.getInstance().isAlgaeDetected()) {
+    if (goToIntake && ScoringSubsystem.getInstance().isCoralDetected()) {
+      setGoToIntake(false);
+    }
+
+    if (goToIntake && ScoringSubsystem.getInstance().isAlgaeDetected()) {
```
jkleiber, line 370: else if please

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -599,10 +601,11 @@ public boolean isDriveLiningUp() {
    * @return true if location is reef; false otherwise (processor / coral station)
    */
   public boolean isDesiredLocationReef() {
-    return !(desiredLocation == DesiredLocation.CoralStationLeft
-            || desiredLocation == DesiredLocation.CoralStationRight
-            || desiredLocation == DesiredLocation.Processor)
-        && !goToIntake; // only want reef if intake is false
+    return isLocationAlgaeIntake(desiredLocation) && goToIntake
```
jkleiber, line 604: Can you add parentheses around this algae condition?

Even better can we split this huge boolean expression into 2: isCoralReefTarget, isAlgaeReefTarget and OR them together 

That would be way more readable

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -612,7 +615,18 @@ public boolean isDesiredLocationReef() {
    */
   public boolean isLocationScoring(DesiredLocation location) {
     return !(location == DesiredLocation.CoralStationLeft
-        || location == DesiredLocation.CoralStationRight);
+            || location == DesiredLocation.CoralStationRight)
+        || !isLocationAlgaeIntake(location);
```
jkleiber, line 619: This logic is now incorrect because everything needs to be OR-ed first and then NOT-ed

If location= coral station right here, then we return true because isLocationAlgaeIntake will be false, not-ed to true, then passed through via the OR

jkleiber (commented):
First pass

src/main/java/frc/robot/subsystems/drive/Drive.java
linglejack06: i separated things into two booleans, then did !isCoralStation && !isAlgaeIntake. LMK if this works

jkleiber:
This looks good to me, we should test on robot before merging though

src/main/java/frc/robot/InitBindings.java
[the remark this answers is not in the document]
aidnem: I think we should start scoring at intake as soon as we enter OTF because then scoring will already be positioned to intake while it drives into the algae.

I think your code looks like it's fine I just commented to confirm that this is how you thought it would work too. I'll read latest changes.

jkleiber:
@linglejack06 I'm good to merge this if we put some sim evidence in the PR

linglejack06:
I'll add that today

jkleiber:
@linglejack06 you could post a link to google drive if it's too big for GitHub to upload (although 9.9MB doesn't sound like much)

linglejack06:
> @linglejack06 you could post a link to google drive if it's too big for GitHub to upload (although 9.9MB doesn't sound like much)

i think its just slow internet on my end

linglejack06:
https://drive.google.com/file/d/1l6p0L0jMm1UmVIB3mtFEcUZXzRgZSujW/view?usp=drive_link

linglejack06:
ok, should be configured so you can view it, lmk if there is issues

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -364,10 +362,11 @@ public void teleopInit() {
...
     // Manually cancel go to intake if we have a gamepiece
-    if (goToIntake && ScoringSubsystem.getInstance().isCoralDetected()
-        || ScoringSubsystem.getInstance().isAlgaeDetected()) {
-      setGoToIntake(false);
-    }
+    // if (goToIntake && ScoringSubsystem.getInstance().isCoralDetected()) {
+    //   setGoToIntake(false);
+    // } else if (goToIntake && ScoringSubsystem.getInstance().isAlgaeDetected()) {
+    //   setGoToIntake(false);
+    // }
```
jkleiber, line 369: why did we comment this out?

linglejack06: didn't mean to, it was part of testing when i had logic error

linglejack06:
@jkleiber this is ready for review

jkleiber (approved):
approved

PR #159, opened 2025-03-06
Intentionally throw algae into the net

aidnem:
Hold the wrist pointed upward and the elevator below max height in warmup for net, then snap wrist out to score positions while running rollers in score to throw the algae into the net.

This also adds a protection clamp to scoring to keep the wrist under a certain angle when there's an algae in the robot and the elevator is below a certain height to avoid slapping the algae into the crossbar or a stage of the elevator.

### DIFF
--- a/src/main/deploy/constants/ScoringSetpoints.json
+++ b/src/main/deploy/constants/ScoringSetpoints.json
@@ -83,7 +83,7 @@
       "unit": "Meter"
     },
     "wristAngle": {
-      "value": -0.105,
+      "value": -0.095,
       "unit": "Rotation"
     }
   },
@@ -94,7 +94,7 @@
       "unit": "Meter"
     },
     "wristAngle": {
-      "value": -0.105,
+      "value": -0.095,
       "unit": "Rotation"
     }
   },
--- a/src/main/deploy/constants/ScoringSetpoints.json
+++ b/src/main/deploy/constants/ScoringSetpoints.json
@@ -123,12 +123,12 @@
   "netWarmup": {
     "name": "netWarmup",
     "elevatorHeight": {
-      "value": 0.25,
+      "value": 1.55,
       "unit": "Meter"
     },
     "wristAngle": {
-      "value": -0.105,
-      "unit": "Rotation"
+      "value": 1.3,
+      "unit": "Radian"
     }
   },
   "net": {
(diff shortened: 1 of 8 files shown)


src/main/deploy/constants/ScoringSetpoints.json
```diff
@@ -123,12 +123,12 @@
   "netWarmup": {
     "name": "netWarmup",
     "elevatorHeight": {
-      "value": 0.25,
+      "value": 1.55,
```
jkleiber, line 126: Just a note: I think this might be too low - the wrist snap needs to happen at the top of the elevator. Obviously test and tune in shop as needed

src/main/java/frc/robot/subsystems/scoring/states/ScoreState.java
```diff
@@ -21,14 +20,8 @@ public ScoreState(ScoringSubsystem scoringSubsystem) {
...
-    if (scoringSubsystem.getGamePiece() == GamePiece.Algae
-        && scoringSubsystem.getTarget() == FieldTarget.Net) {
-      // Force the net setpoint only in score since the "warmup" for the net stays low to use
-      // elevator to shoot algae upward
-      setpoint = JsonConstants.scoringSetpoints.net;
-    } else {
-      setpoint = ScoringSetpoints.getWarmupSetpoint(scoringSubsystem.getTarget());
-    }
+    // Only warmup like normal when we aren't doing algae in the net
+    setpoint = ScoringSetpoints.getWarmupSetpoint(scoringSubsystem.getTarget());
```
jkleiber, line 24: Removing the if statement doesn't make sense to me here? I would picture the warmup to be wrist pulling the algae into the robot and being fairly high. Then when we score we want to flick the wrist toward the net, which is a different setpoint

So wouldn't we want the exact same logic as before?

jkleiber (commented):
some comments on the logic here

src/main/java/frc/robot/subsystems/scoring/states/ScoreState.java
aidnem: This is true, I forgot that I made the warmup setpoint different in getWarmupSetpoint too

jkleiber (approved):
Approved, we can tune the values to make this more consistent on a competition branch

aidnem:
@jkleiber I ended up being unable to come to shop yesterday but I was told that the score button didn't spin rollers when scoring in processor. I've created #162 to track that work but i'm mentioning it here so that both of us remember that there is work that needs to be done before algae is "complete" (we can probably test this in the pit).

PR #160, opened 2025-03-06
adds brake switch for when disabled

linglejack06:
- add brake switch only
- can be merged after test since it doesnt have led stuff

### DIFF
--- a/src/main/java/frc/robot/BuildConstants.java
+++ b/src/main/java/frc/robot/BuildConstants.java
@@ -5,13 +5,13 @@ public final class BuildConstants {
   public static final String MAVEN_GROUP = "";
   public static final String MAVEN_NAME = "2025-Robot-Code";
   public static final String VERSION = "unspecified";
-  public static final int GIT_REVISION = 54;
-  public static final String GIT_SHA = "db9d6b851a6c92c5a76249f543423b79bf76d9e4";
-  public static final String GIT_DATE = "2025-03-06 18:30:13 EST";
+  public static final int GIT_REVISION = 56;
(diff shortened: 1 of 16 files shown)


linglejack06:
@aidnem @jkleiber can someone check logic for brake and be ready to merge once i test?
elevator can also be in brake mode right now, do we want that, or put in coast like it is in main?

jkleiber:
@linglejack06 @aidnem were you guys able to test this?

linglejack06:
I tested the previous commit, in which I verified that configs change. But they changed every single loop. So I fixed that with the next commit, but never got to test it.

The only problem was I wasn't checking the previous value of brake switch before sending it to all of the subsystems. But i added that

jkleiber:
@linglejack06 could you add the LED switch to this as well?

I woild actually be fine merging this after that is done and the format looks good because this should be mostly tested and doesn't impact robot function that much

If it's broken somehow we can also easily disable it

linglejack06:
yeah i can do that, are led merged?

jkleiber:
@linglejack06 I merged LEDs this morning

linglejack06:
added, hopefully format works, i cant build since it needs to download all of the new wpilib packages and i dont have internet well enough

jkleiber (commented):
feature flag comment, also format is failing 😢

linglejack06:
ill try to get format working, vpn is giving me 1kb/sec right now lol

jkleiber (approved):
approved for when format gets working

aidnem:
@linglejack06 I'm leaving this so I don't steal credit for the work by formatting but let me know if you'd rather I just get it done.

PR #163, opened 2025-03-07
Actually run rollers when scoring algae into processor in ScoreState

aidnem:
Previously, the net score logic was incomplete and resulted in not running the rollers when trying to score in processor

- Add an else statement to ScoreState so that the rollers run when trying to score an algae *not* into the net.


aidnem:
This might be such a trivial change that we don't even need to test it before merge. Imo we should merge and then if this somehow doesn't work after review and merge it can get fixed when we do a pit systems check at comp tomorrow morning.

jkleiber (approved):
Go ahead and merge this

PR #173, opened 2025-03-12
add ability to stop lineup by moving joysticks

linglejack06:
- lineup will stop running if any field centric speeds are passed to the robot with greater than 0 velocity (angular, vx, or vy)

### DIFF
--- a/src/main/java/frc/robot/BuildConstants.java
+++ b/src/main/java/frc/robot/BuildConstants.java
@@ -3,14 +3,14 @@ package frc.robot;
 /** Automatically generated file containing build version information. */
 public final class BuildConstants {
   public static final String MAVEN_GROUP = "";
-  public static final String MAVEN_NAME = "2025-Robot-Code-3";
+  public static final String MAVEN_NAME = "2025-Robot-Code";
   public static final String VERSION = "unspecified";
-  public static final int GIT_REVISION = 68;
-  public static final String GIT_SHA = "9899bcb2854cc592c52ca78edd1259dc08feb736";
-  public static final String GIT_DATE = "2025-03-08 09:57:29 EST";
-  public static final String GIT_BRANCH = "156-ramp-work";
-  public static final String BUILD_DATE = "2025-03-08 10:03:36 EST";
-  public static final long BUILD_UNIX_TIME = 1741446216963L;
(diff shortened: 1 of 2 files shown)


linglejack06:
**

https://github.com/user-attachments/assets/e875ca60-6cfa-468d-86d5-9564549e185b

**

linglejack06:
@jkleiber @aidnem aiden mentioned this was something we wanted so i went ahead and added it and attached sim proof so we can merge. pretty simple change

linglejack06:
@aidenm someone review????

jkleiber:
@linglejack06 @aidnem why do we need this? Can't the driver just stop lineup by letting go of the score button?

I'm worried that adding this will result in us accidentally canceling lineup in teleop. I don't recall this being an issue but maybe I need to talk to Oliver about what they want

linglejack06:
I think if we were to stop pressing scote trigger, the elevator would go down right @aidnem 

So this allows over riding lineup while still having ability to manual score. Without having to enter manual mode

jkleiber:
I see, thanks for the explanation 

Would like to test this on the robot before merging tbh. I think the code looks fine though

jkleiber:
In shop testing report: this does not work

The robot no longer aligns and auto scores randomly into the ground. I think that very small joystick movements are kicking us out of lineup and then for some reason that causes a score once the scoring subsystem is in position

We need a cleaner implementation that waits for driver manual score if this scenario happens

linglejack06:
Ok that's weird, just lmk how you'd like it to be setup and I'll update

linglejack06:
> In shop testing report: this does not work
> 
> 
> 
> The robot no longer aligns and auto scores randomly into the ground. I think that very small joystick movements are kicking us out of lineup and then for some reason that causes a score once the scoring subsystem is in position
> 
> 
> 
> We need a cleaner implementation that waits for driver manual score if this scenario happens

Issue with scoring makes sense, drive lined up gets set to true when we exit lineup. Maybe joystick headband is too little as to alignment cancelling without desiring it too

aidnem:
@linglejack06 maybe we could just set is drive lined up to whether or not it's actually lined up on exit instead. Also maybe having a wider dead band on joysticks for leaving lineup would be good.

PR #176, opened 2025-03-12
updating climb code for new design

minhnguyenbhs:
Switches code from cancoder to candi (works on hitl but not tested on robot)!


minhnguyenbhs:
https://github.com/user-attachments/assets/5031b5c2-92b1-4206-a9d8-620ca8e66ead

The angle is reported in rotations, should I make the value mod 1 rotation, or should I just wait for the candi to be put on the robot and just change the offset point then?

minhnguyenbhs:
hi @aidnem !! was this what you were looking for?

src/main/deploy/constants/hitl/ClimbConstants.json
```diff
@@ -11,8 +11,9 @@
     "value": 0.0,
     "unit": "Degree"
   },
-  "leadClimbMotorId": 16,
+  "leadClimbMotorId": 6,
   "followerClimbMotorId": 17,
+  "canDiID": 0,
```
aidnem, line 16: Could you please add this constant to comp/ClimbConstants.json as well?

src/main/java/frc/robot/subsystems/climb/ClimbIO.java
```diff
@@ -36,7 +36,5 @@ public default void setBrakeMode(boolean brake) {}
 
   public default void setPID(double p, double i, double d) {}
 
-  public default void setFF(double newFF) {}
```
aidnem, line 39: Thanks for removing this horrible code lol

src/main/deploy/constants/config.json
```diff
@@ -1,5 +1,5 @@
 {
-  "environment": "comp",
+  "environment": "hitl",
```
aidnem, line 2: This needs to be changed back to comp before it's merged

src/main/java/frc/robot/subsystems/climb/ClimbIOSim.java
```diff
@@ -13,8 +13,9 @@
 public class ClimbIOSim extends ClimbIOTalonFX {
 
   TalonFXSimState leadMotorSimState = leadMotor.getSimState();
-  TalonFXSimState followerMotorSimState = followerMotor.getSimState();
-  CANcoderSimState climbEncoderSimState = climbAngleCoder.getSimState();
+  // TalonFXSimState followerMotorSimState = followerMotor.getSimState();
```
aidnem, line 16: Did this get commented out by accident or is there only 1 motor on climb now?

src/main/java/frc/robot/subsystems/climb/states/IdleState.java
```diff
@@ -13,7 +13,10 @@ public IdleState(ClimbSubsystem climbSubsystem) {
 
   @Override
   public void periodic() {
-    climbSubsystem.setFeedforward(0.0);
+    climbSubsystem.setPID(
+        ClimbConstants.synced.getObject().climbkP,
+        ClimbConstants.synced.getObject().climbkI,
+        ClimbConstants.synced.getObject().climbkD);
```
aidnem, line 19: setting PID in periodic will cause a pretty huge amount of lag (also repeatedly flashing configs is probably not good for the motor's memory). Could this be moved to onEntry instead?

src/main/java/frc/robot/subsystems/climb/states/LiftingState.java
```diff
@@ -15,7 +14,10 @@ public LiftingState(ClimbSubsystem climbSubsystem) {
 
   @Override
   public void periodic() {
-    climbSubsystem.setFeedforward(JsonConstants.climbConstants.climbFFVolts);
+    climbSubsystem.setPID(
```
aidnem, line 17: Same as my other comment, this should probably be in onEntry

src/main/java/frc/robot/subsystems/climb/states/SearchingState.java
```diff
@@ -14,7 +14,10 @@ public SearchingState(ClimbSubsystem climbSubsystem) {
 
   @Override
   public void periodic() {
-    climbSubsystem.setFeedforward(0.0);
+    climbSubsystem.setPID(
```
aidnem, line 17: See other 2 comments

src/main/java/frc/robot/subsystems/climb/states/WaitingState.java
```diff
@@ -14,7 +14,10 @@ public WaitingState(ClimbSubsystem climbSubsystem) {
 
   @Override
   public void periodic() {
-    climbSubsystem.setFeedforward(0.0);
+    climbSubsystem.setPID(
```
aidnem, line 17: Same as other comments, move to onEntry

aidnem (changes requested):
This looks good, just a few questions/change requests

src/main/deploy/constants/comp/ClimbConstants.json
```diff
@@ -1,18 +1,19 @@
...
     "unit": "Radian"
   },
   "searchingAngle": {
-    "value": 1.1,
+    "value": 0.6,
     "unit": "Radian"
   },
   "finalHangingAngle": {
-    "value": 0.4,
+    "value": -0.5,
```
aidnem, line 11: Hopefully the encoder was zeroed before finding these setpoints. This could be verified tomorrow at shop though too

src/main/deploy/constants/comp/FeatureFlags.json
```diff
@@ -1,7 +1,7 @@
 {
-  "runDrive": true,
-  "runVision": true,
-  "runScoring": true,
+  "runDrive": false,
+  "runVision": false,
+  "runScoring": false,
   "runClimb": true,
-  "runRamp": true
+  "runRamp": false
```
aidnem, line 6: Make sure these get reenabled before merge :)

aidnem (commented):
This looks good to merge after featureflags get turned back on (and it's tested on robot)

jkleiber:
After Preston and I spent an hour on trying to understand how the candi + rev encoder combo actually works, we gained some insights but ultimately this configuration does not consistently produce the same position values

The issue is that despite using an absolute encoder, the startup conditions matter for what the value is. Therefore, the plan is to switch to a cancoder similar to what we did for highkey shooter. There will be a setup for this by shop tomorrow most likely, so we should get this branch set up for that. 

The climb worked well whenever we did a Phoenix tuner manual test of it, so we just need control of the arm

aidnem:
@jkleiber if we go back to CANcoder that would mean pretty much reverting this PR/closing and then zeroing encoder and getting new setpoints for climb right?

jkleiber:
@aidnem that's correct. However we should keep this open for if the cancoder solution doesn't work out and we end up having to write some super sketchy code to determine where the encoder was at boot

src/main/java/frc/robot/subsystems/climb/ClimbIOTalonFX.java
```diff
@@ -104,7 +105,10 @@ public ClimbIOTalonFX() {
 
     // TODO: set lockedToCage when ramp becomes available
   }
-
+  @Override
+  public void robotInit() {
```
aidnem, line 109: Just realized we never actually call the robotInit method on this IO. That means this change won't do anything if we test it in current state. @j0shuaS can you add calls to this before shop or do you want me to take care of it? (you'll have to go through Robot.robotInit -> RobotContainer.robotinit -> ClimbSubsystem.robotInit -> IO.robotInit)

jkleiber:
closing this PR since #200 is the correct one now that we have the cancoder back

Issue #120, opened 2025-02-22
Vary wrist setpoint based on along track distance


aidnem:
**Purpose**
Scoring needs to have different setpoints for scoring when not touching the reef (auto & mixed modes) vs touching the reef (manual)

**Project Scope**
- [ ] Add code to switch which setpoints smartly
- [ ] find the setpoints for both

jkleiber:
@aidnem I recommend closing this since aiming for a precise point in space results in limit cycles. Can revisit in the future if we have a specific need to do this for L1

aidnem:
@jkleiber Before I close this, I had one potential solution (for instance in case a coral gets stuck in front of the reef in auto, which would block us from ever scoring): If robot detects that along track is too large and not coming down, it could fallback to the far setpoints. We'd ideally never use them but it could be useful for unlucky times.

If we do go this route it would be pretty low priority since I'd rather be actually able to score than waste our time doing this and then not be able to score close or far away.

jkleiber:
Let's keep this issue open then to establish some lookup table based on along track/cross track distance 

It may be overkill but it could be something that opens up our tolerance on drive, so we should explore it if we have time in the future

jkleiber:
@aidnem @linglejack06 I've renamed this ticket and reprioritized it to Blacksburg comp. I think if we can create a lookup table of required wrist position based on along track distance we can go even faster in auto and teleop

I don't think we'll need to factor in robot velocity, but we will see when we actually test this on the robot

The way I see this working is some JSON lookup table (2 lists, one of along-track distance, another of wrist setpoints). Then we use interpolate double in coppercore to pick the wrist setpoint based on vision along track distance. We should have a feature flag for this, and maintain the constant wrist setpoint as a backup since we know it works at comp.

jkleiber:
Yeah we can see how it goes. If we can keep it a fixed height that is preferable though because it's simpler. I'm not sure how much height variation will buy us, but it might be worth drawing it out since this is a geometry problem

jkleiber:
Our reef at home is likely not accurate enough to do this tuning, and we are running up against time constraints for VABLA. Moving this to DCMP milestone to reduce schedule pressure

jkleiber:
wIth the new claw setup, this ticket is no longer relevant - closing

PR #177, opened 2025-03-13
(#120) Add Distance based wrist setpoints

aidnem:
ScoringSetpoints now uses drive lineup's alongtrack distance to pick an angle for the wrist using InterpolateDouble


aidnem:
**Remaining work:**
- Decide if this needs elevator to move too
- Test in sim (setpoints can be determined whenever robot is available at shop, possible after merge)

src/main/java/frc/robot/constants/JsonConstants.java
```diff
@@ -43,6 +43,15 @@ public static void loadConstants() {
...
     drivetrainConstants = DrivetrainConstants.synced.getObject();
     visionConstants = VisionConstants.synced.getObject();
+
+    try {
+      scoringSetpoints.fillVariableL4WristMap();
+    } catch (Exception e) {
+      e.printStackTrace();
+      System.out.println(
+          "WARNING: L4 variable wrist map constants were incorrectly defined (different numbers of distances vs. angles), using L4 setpoint as placeholder map");
+      scoringSetpoints.fillPlaceholderL4WristMap();
```
jkleiber, line 53: If the map is defined incorrectly, I think the appropriate action is to turn off variable scoring rather than filling in the table. Reasoning: we can be in position with the wrist but not the correct distance away for the nominal setpoint if we fill the table out, and then we'll be missing shots all over the place

src/main/java/frc/robot/constants/ScoringSetpoints.java
```diff
@@ -66,12 +69,88 @@ public static record ScoringSetpoint(String name, Distance elevatorHeight, Angle
...
+   * Combined with variableL4WristAngles, these form the interpolate double used for variable
+   * distance L4 scoring
+   */
+  public final Distance[] variableL4AlongTrackDistances = {Meters.of(0.1), Meters.of(0.0)};
+
+  /**
+   * Combined with variableL4AlongTrackDistances, these form the interpolate double used for
+   * variable distance L4 scoring
+   */
+  public final Angle[] variableL4WristAngles = {Rotations.of(0.32), Rotations.of(0.09)};
```
jkleiber, line 82: Are these / other constants added here placeholders, or are they derived from something?

src/main/java/frc/robot/constants/ScoringSetpoints.java
```diff
@@ -66,12 +69,88 @@ public static record ScoringSetpoint(String name, Distance elevatorHeight, Angle
...
+   * called AFTER drivetrain constants are synced using JSON (it depends on a synced JSON value from
+   * drivetrainConstants)
+   */
+  public void fillPlaceholderL4WristMap() {
+    HashMap<Double, Double> wristMap = new HashMap<>();
+
+    wristMap.put(0.0, L4.wristAngle().in(Rotations));
+
+    variableL4WristMap = new InterpolateDouble(wristMap);
+  }
```
jkleiber, line 139: See other comment about not failing into a placeholder, but rather the original single setpoint behavior

src/main/java/frc/robot/subsystems/drive/states/LineupState.java
jkleiber, line 1: I don't know where it belongs (maybe this file) but we need to make sure the allowed along track distance to score at is the same as the maximum distance in the table. For example:

```
if (alongTrackDistance < tableMaximumDistance && crossTrackError < tolerance ... ) {
// command score
}
```
This alongtrack maximum will depend on if we are using the table or if we are in single setpoint mode

jkleiber (commented):
first pass review to give some early design feedback

src/main/java/frc/robot/constants/ScoringSetpoints.java
aidnem: The closest setpoint is the current L4 setpoint while the furthest one is the current idle setpoint (I assumed that angle will be close to how we'll end up scoring)

src/main/java/frc/robot/constants/ScoringSetpoints.java
```diff
@@ -66,12 +69,99 @@ public static record ScoringSetpoint(String name, Distance elevatorHeight, Angle
...
+      wristMap.put(distanceMeters, variableL4WristAngles[i].in(Rotations));
+
+      if (distanceMeters > maxVariableDistance) {
+        maxVariableDistance = distanceMeters;
+      }
+    }
+
+    variableL4WristMap = new InterpolateDouble(wristMap);
+
+    usingVariableL4 = true;
```
jkleiber, line 140: Not sure where, but we should log this in scope

jkleiber (commented):
Just one nit

I also want to take this to the robot and tune the values before merge

aidnem:
All comments are resolved, I'm going to write out JSON constants for this and then it should be ready to test tomorrow

jkleiber:
Closing this PR since it isn't possible in the current robot now that the carriage has been moved further in

PR #179, opened 2025-03-13
Bouncy ramp

avidraccoon:
Add new state to ramp state machine to make the ramp bounce to extend new mechanism on the ramp


src/main/java/frc/robot/subsystems/ramp/RampSubsystem.java
```diff
@@ -67,32 +68,35 @@ private void setupStateMachine(RampMechanism mechanism) {
...
+    config.configure(RampStates.HOMING).permit(RampTriggers.HOMING_FINISHED, RampStates.IDLE);
 
     config
-        .configure(RampStates.CLIMB)
-        .permit(RampTriggers.GOTO_IDLE, RampStates.IDLE)
-        .permit(RampTriggers.START_INTAKE, RampStates.CLIMB_TO_INTAKE_HOMING);
+        .configure(RampStates.STARTUP_HOMING)
+        .permit(RampTriggers.HOMING_FINISHED, RampStates.EXTENDING);
+
+    config.configure(RampStates.EXTENDING).permit(RampTriggers.HOME, RampStates.HOMING);
```
jkleiber, line 91: Extending needs to be able to go to idle rather than creating an infinite loop with homing

src/main/java/frc/robot/subsystems/ramp/RampSubsystem.java
jkleiber, line 1: There are way too many unrelated changes in this file. Please undo the unrelated changes so the review is easier

src/main/java/frc/robot/subsystems/ramp/states/ExtendingState.java
```diff
@@ -0,0 +1,22 @@
...
+import frc.robot.subsystems.ramp.states.RampState.RampTriggers;
+
+public class ExtendingState extends RampState {
+
+  @Override
+  public void periodic() {
+    if (mechanism.inputs.position < JsonConstants.rampConstants.extendPosition) {
+      setVoltage(JsonConstants.rampConstants.extendingVoltage);
+    } else {
+      fireTrigger.accept(RampTriggers.HOME);
```
jkleiber, line 13: This should go to Idle instead right?

src/main/java/frc/robot/subsystems/ramp/states/ExtendingState.java
```diff
@@ -0,0 +1,22 @@
...
+    }
+    super.periodic();
+  }
+
+  @Override
+  public boolean inPosition() {
```
jkleiber, line 19: I think if we are in this state we are never in position to do anything, so we should just return false

jkleiber (commented):
First pass review. Please remove unrelated changes to make the code diff more focused on changes associated with the issue

src/main/java/frc/robot/subsystems/ramp/RampSubsystem.java
avidraccoon: It doesn't create an infinite loop because there are multiple states based on the Homing class. This goes to the homing that returns to idle.

src/main/java/frc/robot/subsystems/ramp/states/ExtendingState.java
avidraccoon: This is done to ensure no loss of position information. It should be the same speed

avidraccoon:
@jkleiber I can see two ways of removing the unrelated changes.
1. I redo this entire branch
2. I make a new branch based on this one remove the bouncy ramp changes. Make a new pr and merge that one. Then handle merge changes on this one and merge.

src/main/java/frc/robot/subsystems/ramp/RampSubsystem.java
jkleiber: Oh I see, I was missing the distinction between startup_homing and homing

jkleiber: That being said, it looks like this PR deletes the configuration that sends homing to idle. We probably should fix that

jkleiber:
> @jkleiber I can see two ways of removing the unrelated changes.
> 1. I redo this entire branch
> 2. I make a new branch based on this one remove the bouncy ramp changes. Make a new pr and merge that one. Then handle merge changes on this one and merge.

I think I have a better grasp of this PR now, so just a note for future reference: just make the minimum changes unless a drive by fix is really needed. Oftentimes less is more

src/main/java/frc/robot/subsystems/ramp/RampSubsystem.java
avidraccoon: it is still there just moved

jkleiber (approved):
Approved

src/main/deploy/constants/comp/FeatureFlags.json
```diff
@@ -1,7 +1,7 @@
 {
-  "runDrive": true,
-  "runVision": true,
-  "runScoring": true,
-  "runClimb": true,
+  "runDrive": false,
+  "runVision": false,
+  "runScoring": false,
+  "runClimb": false,
```
jkleiber, line 5: these need to be set back to true before we merge

jkleiber (approved):

PR #182, opened 2025-03-14
add sidestepping to lineup

linglejack06:


linglejack06:
@jkleiber does this button binding work (right joystick right button)? it will just swap to the other pole.

ill test this in sim some to see if it works but wanted to check button binding with you

jkleiber:
That seems reasonable to me @linglejack06

linglejack06:
got it, not working in sim right now because snakescreen is always publishing a location, so when i attempt to change it from drive, it immediately gets overridden by strategy maanger asking drive to update from network tables

if this feature is only used in the smart mode, then we can just not call the snakescreen thing in that mode, however, if we want this feature in all modes i guess i could just check to see if snakescreen value has changed (only then would we make drive update from there)

do you have a preference? @jkleiber

jkleiber:
@linglejack06 let's start by only making this available in smart mode. I think @aidnem ignores snakescreen in #181 in smart mode, so maybe we just need to merge that for you to get those benefits

linglejack06:
> @linglejack06 let's start by only making this available in smart mode. I think @aidnem ignores snakescreen in #181 in smart mode, so maybe we just need to merge that for you to get those benefits

yeah was about to merge and that should fix it

aidnem:
@linglejack06 smart mode is merged now, rebase/merge those changes in whenever you want and this should work in sim. SnakeScreen doesn't have a smart mode yet, but I just added it myself by adding another entry to src/Autonomy.tsx:
```ts
  const modes: autonomyLevels[] = [
    { value: "high", label: "Full Auto" },
    { value: "smart", label: "Smart" },
    { value: "mid", label: "Teleop" },
    { value: "low", label: "Manual" },
  ];
```

The string value for Smart mode is just "smart"

linglejack06:
https://github.com/user-attachments/assets/146bcfac-94a8-4846-86a2-132215bb5dcc

linglejack06:
@jkleiber sidestepping works now, think it just needs pid tuning still for when we switch cameras, ill add that now

src/main/deploy/constants/comp/DrivetrainConstants.json
```diff
@@ -24,6 +24,9 @@
   "driveCrossTrackKp": 5.0,
   "driveCrossTrackKi": 0.5,
   "driveCrossTrackKd": 0.0,
+  "driveCrossTrackOtherCameraKp": 5.0,
+  "driveCrossTrackOtherCameraKi": 0.5,
+  "driveCrossTrackOtherCameraKd": 0.0,
```
jkleiber, line 29: These gains are almost certainly wrong - making a note here so we tune this in shop before merging

jkleiber:
@linglejack06 this looks good, we should test this today as well as #183

jkleiber:
verdict: this works. I will go ahead and merge

jkleiber (approved):

linglejack06:
@jkleiber did side stepping work too? or just the pid stuff?

jkleiber:
@linglejack06 sidestep worked well enough to merge. It's unclear if the feature itself will be very useful but we'll see with additional driver practice. Thanks for developing it

PR #183, opened 2025-03-14
add 12 individual otf poses

linglejack06:
- add constants
- TODO: make the poses actually different (in line with tag for each camera)


linglejack06:
@jkleiber poses added from pathplanner

jkleiber:
@linglejack06 looks good, we should test this today

jkleiber:
@aidnem FYI: in order to actually get reef level selection for coral to work, we had to undo your stuff for algae setting. I think we need something a bit better, so we opened #187

jkleiber (approved):

PR #185, opened 2025-03-15
improve teleop otf

linglejack06:
- slower speed / acceleration
- dont allow scoring far warmup

### DIFF
--- a/src/main/java/frc/robot/subsystems/drive/states/OTFState.java
+++ b/src/main/java/frc/robot/subsystems/drive/states/OTFState.java
@@ -6,6 +6,7 @@ import com.pathplanner.lib.path.PathConstraints;
 import coppercore.controls.state_machine.state.PeriodicStateInterface;
 import coppercore.controls.state_machine.transition.Transition;
 import edu.wpi.first.math.geometry.Pose2d;
+import edu.wpi.first.wpilibj.DriverStation;
 import edu.wpi.first.wpilibj2.command.Command;
 import frc.robot.constants.JsonConstants;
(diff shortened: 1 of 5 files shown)


src/main/java/frc/robot/subsystems/drive/states/OTFState.java
```diff
@@ -186,7 +196,7 @@ public void periodic() {
 
     if (drive.isDriveCloseForWarmup() && ScoringSubsystem.getInstance() != null) {
       ScoringSubsystem.getInstance().fireTrigger(ScoringTrigger.StartWarmup);
-    } else if (drive.isDriveCloseForFarWarmup() && ScoringSubsystem.getInstance() != null) {
+    } else if (drive.isDriveCloseForFarWarmup() && ScoringSubsystem.getInstance() != null && DriverStation.isAutonomous()) {
```
jkleiber, line 199: Rather than preventing this in teleop, what if we had a shorter far warmup distance in teleop?

I agree that we don't want to get top-heavy under defense, but I think far warmup is nice for getting ready to score quicker than if we wait. I don't want to slow down cycle times too much

linglejack06: ok ill add that

linglejack06:
@jkleiber far warmup should happen at 2 meters now in teleop (rather than 2.5) @aidnem is that fine?

jkleiber:
just noting here that this seems testable in the current form. That being said, I think this is lower on the priority queue than auto testing and climb (just an FYI)

PR #190, opened 2025-03-15
Fixes a JSON classic issue that was not allowing signal strength to be set

jkleiber:
see title


aidnem (approved):
We got JSON classicced again 😭. thanks for fixing

PR #192, opened 2025-03-16
Actually use snakescreen scoring level and Smart mode intake location in Smart autonomy

### LINKED ISSUE #187: Ignoring Snake Screen scoring level bad in smart autonomy mode
When we update scoring level on snake screen when in auto it does not update.

### LINKED ISSUE #189: Smart autonomy intake selection doesn't work
## Summary

When doing intake in smart autonomy, the snakescreen value is overriding the automatic pick

aidnem:
- Don't update intake location from snakescreen when in smart mode; this is done when intake is pressed based on closest distance (closes #189)
- Update scoring level from snakescreen when warmup is pressed; this means you should be able to select reef level, instead of it being ignored like it was before (closes #187)

### DIFF
--- a/src/main/java/frc/robot/StrategyManager.java
+++ b/src/main/java/frc/robot/StrategyManager.java
@@ -228,19 +228,20 @@ public class StrategyManager {
   public void updateScoringLocationsFromSnakeScreen() {
     // drive reef location
     if (drive != null) {
-      // Don't set reef target from SnakeScreen in 'smart' mode, this will be picked automatically
+      // Don't set reef target or intake location from SnakeScreen in 'smart' mode, this will be
+      // picked automatically
       // based on distance
       if (getAutonomyMode() != AutonomyMode.Smart) {
         drive.updateDesiredLocationFromNetworkTables(
             reefLocationSelector.get(), gamePieceSelector.get().equalsIgnoreCase("algae"));
-      }
 
-      // 20: left; 21: right
-      if (gamePieceSelector.get().equalsIgnoreCase("coral")) {
-        drive.setDesiredIntakeLocation(
-            intakeLocationSelector.get() == 20
-                ? DesiredLocation.CoralStationLeft
-                : DesiredLocation.CoralStationRight);
+        // 20: left; 21: right
+        if (gamePieceSelector.get().equalsIgnoreCase("coral")) {
+          drive.setDesiredIntakeLocation(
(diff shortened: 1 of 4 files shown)


aidnem:
Testing this now and then it'll be ready for review

aidnem:
Coral levels work but intake location is still being suspicious, looking into it further

src/main/java/frc/robot/StrategyManager.java
```diff
@@ -255,12 +256,14 @@ public void updateScoringLocationsFromSnakeScreen() {
...
-      // Don't automatically set level in smart mode; this will be set when the warmup trigger is
+      // Don't automatically set level in smart mode FOR ALGAE; this will be set when the warmup
+      // trigger is
       // pressed. This can't happen in periodic because algae level is automatically determined when
       // intake is pressed and would be overridden by this in each loop.
-      // if (getAutonomyMode() != AutonomyMode.Smart) {
-      scoringSubsystem.updateScoringLevelFromNetworkTables(reefLevelSelector.get());
-      // }
+      if (getAutonomyMode() != AutonomyMode.Smart
+          || scoringSubsystem.getGamePiece() != GamePiece.Algae) {
```
jkleiber, line 264: We will want to be able to choose net vs processor

aidnem: Whenever you enter warmup, it selects level from SnakeScreen so you should be able to. It bypasses that method and just directly sets the scoring level (it calls the method that's within that if statement). Technically I suppose we don't need it at all (we should just never update it when in smart autonomy) but I figured we might as well have it be correct all the time for coral.

jkleiber: Sounds good

aidnem:
Here's it driving to the nearest coral station (the pauses are me hitting the wrong button lol)

https://github.com/user-attachments/assets/394b6c2c-024e-453a-94a7-deb8e58ed6bd

aidnem:
Algae OTF doesn't work at all and neither does intaking algae in sim, I'll try to fix them so I can test the scoring level stuff for algae too

aidnem:
Algae intake now works too, although net OTF doesnt.

aidnem:
This should be ready to merge (or test on real robot if you want) but I verified it as working in sim

jkleiber (approved):
Approved. Sim evidence is sufficient for high level logic like this

PR #193, opened 2025-03-17
add pre load of otf command

linglejack06:


linglejack06:
@aidnem @jkleiber heres the branch for fixing auto stuff, check out the otf implementation justin and see if thats what you wanted

src/main/java/frc/robot/subsystems/scoring/states/ScoreState.java
```diff
@@ -18,6 +21,11 @@ public ScoreState(ScoringSubsystem scoringSubsystem) {
     this.scoringSubsystem = scoringSubsystem;
   }
 
+  @Override
+  public void onEntry(Transition transition) {
+    OTFState.getDriveToPoseCommand(Drive.getInstance());
```
jkleiber, line 26: Since this is actually running the OTF warmup, the lagspike will occur before we apply outputs. We likely will need to do @aidnem 's suggestion from the ticket yesterday if this is the implementation strategy 

(The suggestion as I see it is add a check in scoring subsystem periodic for if it's our first time running the claw rollers in score state, and then do the OTF logic)

linglejack06: updated

linglejack06:
@aidnem feel free to add the next pose thing however you want, probably strategy manager would be the best option (just use peek on the queue which will return next action without removing from queue then send it to otf state as the pose is set uo as static now)

DONT send it to drive as that will break lineup when it thinkts it should align somewhere else

aidnem:
@linglejack06 I'll add the next location peek and see if it helps performance. I'm also going to test this vs. main to see if there's any reproducible lag in sim

aidnem:
I'm going to test this in sim, hopefully I didn't totally cook OTF in auto

aidnem:
I *think* I'm done with logic right now. After shop when we've confirmed that just instantiating the command causes the pathfind to happen, I can clean the logic up a little bit more, I've left it in a state where it would be safe to add a pathfindcommand.execute call to the drive OTF warmup and not cause issues.

src/main/java/frc/robot/subsystems/drive/states/OTFState.java
```diff
@@ -258,35 +298,38 @@ public void periodic() {
...
+    driveToPose =
+        getDriveToPoseCommand(
+            nextAction.location(),
+            nextAction.type() == ActionType.Intake,
+            Drive.getInstance().isAllianceRed());
+
+    System.out.println(driveToPose == null);
+    if (driveToPose == null) {
+      Drive.getInstance().fireTrigger(DriveTrigger.CancelOTF);
+    }
```
jkleiber, line 329: Do we actually want to cancel OTF in this scenario? I thought we were just trying over and over until it works?

aidnem: driveToPose would only be null here if the OTF pose is null, which would indicate that the next desiredlocation was invalid I think. Also, we wouldn't actually be in OTF state here anyway since we're warming up while still in lineup state. I'll delete this.

jkleiber: Oh nvm, that's only if it randomly disappears during periodic. For now I think this is fine

aidnem: In any case this doesn't need to be checked here since under warmupForNextLocation we should never be in OTF anyway

aidnem:
Sounds good. Hopefully it's noticeably faster. When I tested in sim there were definitely some spikes during Score state instead of at the start of OTF but sim was weird so I'm not 100% convinced.

src/main/java/frc/robot/subsystems/scoring/states/IntakeState.java
```diff
@@ -42,6 +43,10 @@ public void onEntry(Transition transition) {
     detectedAngle = Rotations.mutable(0.0);
     isCounting = false;
     Logger.recordOutput("claw/intake/diff", 0.0);
+
+    if (DriverStation.isAutonomous()) {
+      scoringSubsystem.setShouldWarmupNextOTF(true);
+    }
```
jkleiber, line 49: Quick question @aidnem - when does this intake state start during auto? Is it after OTF finishes?

My concern is that we could create a conflict here between the drive to intake OTF and the drive to reef OTF

aidnem: @jkleiber yes it will. I'll see if drive ever leaves OTF when intaking. If it does, I'll trigger the next warmup when leaving OTF.

If not, we have a couple of options:

1. Don't warmup OTF for intake (lame but easy option)
2. Make OTF state keep track of its next OTF command. In OTF onEntry, we can check if the next planned OTF command has the correct DesiredLocation, and if it is correct we move planned OTF command to current OTF command and set planned command to null.

aidnem: Looks like it will leave OTF when in take.

```java
      } else if (driveToPose.isFinished()) {
        drive.fireTrigger(DriveTrigger.CancelOTF);
        System.out.println("driveToPose isFinished canceled OTF!");
      }
```

I'll add a trigger here to warm up the next OTF command. My only concern is if our HP somehow locks in insanely much and we overshoot the coral by intaking in less than 0.2 seconds but I don't see that happening

jkleiber: What would it take to get the CANRange hooked up as a limit switch on the claw motor? That way it wouldn't even be subject to that overshoot.

Currently HP reaction time is a bit under half second per intake, but that's starting from when we are close enough (in my judgment) to intake rather than when the robot stops moving. That said, I don't think HP + ramp time is ever going under 0.2 seconds

aidnem: No idea what it would take to hook up CANrange as a limit switch, I tried configuring it as a limit switch before and I don't think it did anything. I can try it again tonight at shop.

aidnem:
@linglejack06 @jkleiber just realized we didn't actually schedule the new OTF command that it was generating for warmup so it was generating a new one in onEntry anyway as far as I could tell. I just updated code to make it schedule the command if a correct command already exists instead of creating a new one in onEntry so that might end up speeding up our autos as the optimization was intended to.

aidnem:
Also based on the video of a successful 3 piece I think we need to do far warmup earlier (which we already did) and then make early warmup start a little earlier too, we're still waiting on scoring to reach its setpoint.

linglejack06:
> @linglejack06 @jkleiber just realized we didn't actually schedule the new OTF command that it was generating for warmup so it was generating a new one in onEntry anyway as far as I could tell. I just updated code to make it schedule the command if a correct command already exists instead of creating a new one in onEntry so that might end up speeding up our autos as the optimization was intended to.

lol that makes sense

PR #196, opened 2025-03-19
Cancelling OTF when close to coral station

minhnguyenbhs:
hi!!! I was a bit confused about what this ticket was looking for; it seems like it could just be resolved with an if statement, so I added one, but it's probably more complicated than that. What else should I add?


src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -389,6 +390,13 @@ public void periodic() {
       setGoToIntake(false);
     } else if (goToIntake && ScoringSubsystem.getInstance().isAlgaeDetected()) {
       setGoToIntake(false);
+    } else if (goToIntake
+        && ModeConstants.currentMode == Mode.REAL
+        && this.getPose()
+                .getTranslation()
+                .getDistance(OTFState.findOTFPoseFromDesiredLocation(this).getTranslation())
+            < 1) {
+      setGoToIntake(false);
```
jkleiber, line 399: @minhnguyenbhs could we make the "1" a configurable JSON constant. Also, can we add an AND that we need to be in teleop?

If we do these two things we may end up adding this on wednesday. I want to have this as something we can have at our disposal for DCMP if possible since the change is simple

jkleiber: @aidnem ^ can you do this? It was a problem in today's driver practice

src/main/java/frc/robot/constants/subsystems/DrivetrainConstants.java
```diff
@@ -106,6 +106,9 @@ public class DrivetrainConstants {
   public final Double otfPoseDistanceLimit = 0.1;
   public final Double otfPoseEndingVelocity = 0.5;
 
+  / ** When within this distance of the OTF pose in teleop, drive will not OTF to intake */
```
jkleiber, line 109: you're failing format because this comment is spaced incorrectly smh

aidnem: I wrote this on my chromebook and I couldn't run spotlessApply because I had the wrong java version in my github codespace :(

jkleiber (dismissed):
@aidnem check this and merge if you like it. I made a couple changes (see recent commit)

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -397,12 +398,19 @@ public void teleopInit() {
   @Override
   public void periodic() {
     // Manually cancel go to intake if we have a gamepiece
-    if (ScoringSubsystem.getInstance() != null) {
```
aidnem, line 400: This must be checked to avoid NPE if we ever FeatureFlag out scoring

aidnem (dismissed):

jkleiber (approved):

PR #209, opened 2025-03-25
Remove algae current detection

aidnem:
Algae current detection is no longer relevant since we acquired a CANrange for algae. Now all it does is introduce unnecessary problems & complexity into the code.

This PR:
- Removes algae current detection
- Cleans up algae score logic

Work remaining:
- [ ] Test this in sim to make sure it behaves normally
- [ ] Test this on the robot whenever possible.


jkleiber:
This looks good for robot testing. The likely scenario is that we need to adjust the signal strength threshold for algae since it was arbitrarily set to coral (15000).

jkleiber (approved):

Issue #203, opened 2025-03-24
Linear point to point driving


jkleiber:
# Summary 

In auto it might be faster if we made our own motion profiled point to point "linear drive"

We should be able to leverage the WPILib trapezoidal profile here. 

The concept is to command a velocity vector that points from robot position to a goal pose. The velocity magnitude would be based on distance and trapezoidal profile

This would be mainly used in auto as an alternative to OTF to increase driving stability

jkleiber:
@aidnem @minhnguyenbhs 

Thought about this on the ride home: what if we split this into 2 phases

1. Before we are within X degrees of the target position (with respect to the desired heading at the end) drive to a point Y meters behind the target point (projected backwards along target heading)
2. Once within X degrees, drive to the target point

I think this will make it so we can have an ending velocity while smoothly entering lineup. Thoughts?

aidnem:
@jkleiber this sounds like it might work. My only concern is that it will overshoot its target heading (I'm imagining like a pizza slice shaped area in which it can actually drive to the correct target pose) and have to correct back the other way, or that maybe we won't be able to accelerate as fast as we need to be able to in order to change direction instantly like that.

I also thought about this for a while and realized, I don't think we need a PID controller for translation (and maybe rotation as well but I'm not as sure about that one). We know our acceleration, max velocity, target position, etc. Drive already does PID control to take a set of goal speeds and control the drivetrain to those speeds. The way I see it, we should just use a trapezoidal profile: calculate it based on distance to target with zero as our goal, then use that target velocity to generate goal speeds that have direction pointed toward wherever we're going (or behind it if we do the 2 phase system) and a magnitude of the velocity we get from trapezoid profile. Do you foresee any problems with this? (you too 
@minhnguyenbhs)

jkleiber:
I think we don't need a PID either as long as we update a trapezoidal profile each loop

I'm not worried about acceleration in the 2 phase system because we'd just need to make sure we pick good values for the slice angle and length. To make sure we don't get cooked by the thin part of the slice, we could ignore the slice logic and just drive to the final target once we're within Y meters of the final target

jkleiber:
We will need PID for the rotation though imo

PR #215, opened 2025-03-27
Mentor built linear drive

jkleiber:
@aidnem @minhnguyenbhs @linglejack06 see this for a concept on the phase 1/2 stuff I was talking about in #203 . I would recommend testing this in sim first to see what you think, and then maybe test it in shop tomorrow after combining it with some other fixes

Feel free to take this branch and edit it a bunch, this is sort of a starting place


aidnem:
@jkleiber did you try this with no PID and only motion profile? If not, I'll try it and then if it works I'll commit it and add it here.

jkleiber:
> @jkleiber did you try this with no PID and only motion profile? If not, I'll try it and then if it works I'll commit it and add it here.

I briefly tried, but realized that the trapezoidal profile relies on time or an accurate measure of robot velocity to work the way I imagined. So I just went back to profiled PID 

I think profiled PID will actually be fine if we use a bigger distance threshold and tune the P gain higher (so it drives faster). I need to read the implementation, but I'm pretty sure it just limits the PID output based on the constraints, so we could make the gain quite high and it basically becomes a feedforward. But there's probably some benefit to having feedback so I wouldn't go too crazy with it

aidnem:
@jkleiber profiledPIDController just runs your input through a trapezoidal profile and then that profile setpoint through a PID controller. This means the current path is Position -> Position capped to velocity/acceleration -> Output velocity based on position error. If PID just works then we can still use it but I thought it would be simpler and more correct to go Position -> Velocity based on velocity acceleration -> output to goal speeds.

src/main/java/frc/robot/subsystems/drive/states/LinearDriveState.java
```diff
@@ -0,0 +1,300 @@
...
+  private double kDriveToPointTranslationI =
+      DrivetrainConstants.synced.getObject().kDriveToPointTranslationI;
+  private double kDriveToPointTranslationD =
+      DrivetrainConstants.synced.getObject().kDriveToPointTranslationD;
+  private double kDriveTranslationMaxVelocity =
+      DrivetrainConstants.synced.getObject().kDriveTranslationMaxVelocity;
+  private double kDriveTranslationMaxAcceleration =
+      DrivetrainConstants.synced.getObject().kDriveTranslationMaxAcceleration;
+  private double kPositionTolerance = DrivetrainConstants.synced.getObject().kPositionTolerance;
+  private double kVelocityTolerance = DrivetrainConstants.synced.getObject().kVelocityTolerance;
```
aidnem, line 34: Won't these get evaluated on code init and therefore always be the default values?

jkleiber: Yes. This whole section is just based on Minh's PR. I also added some non JSON comments too. We need to clean this whole thing up imo

aidnem:
After over an hour of trying to use an actual trapezoid profile, I think we just go with PID because it'll be easier to get working

jkleiber:
> After over an hour of trying to use an actual trapezoid profile, I think we just go with PID because it'll be easier to get working

This was me last night

Can you go ahead and fix the constants mess we've made? I think this should be testable tonight.

Sim does this weird thing where lineup does wrong things but that shouldn't have changed. We also probably need to incorporate the PID resets we added yesterday

aidnem:
I've pushed my changes to the `trapezoidal-linear-drive` branch and created #216 in case you want to look over it.

aidnem:
I'll fix the stuff on this branch first but here's another insane hear me out: what if we use raw trapezoid profile but just rely on precious setpoint velocity. This would behave more like a path and require no velocity measurement. We would just be trusting drive to keep up reasonably well with acceleration demands of the profile.

jkleiber:
@aidnem yeah I definitely thought about doing that. I think that's basically what ProfiledPID would do with the idea we had earlier to just use the velocity value

src/main/java/frc/robot/subsystems/drive/states/LinearDriveState.java
minhnguyenbhs: oh wait
should the json synced stuff be in the subsystem declaration then
sorry i didn't realize if you're not working on this atm i can change that in a bit

aidnem:
@jkleiber @linglejack06 do you recall making any lineup changes? It does the weird drive to the side behavior again.

jkleiber:
@aidnem I don't think we made any changes but I've seen that in sim so maybe it's a real thing

jkleiber:
@aidnem a couple ideas:

1. We could remove adjustment again - those changes aren't on this branch

2. You could checkout main and see if this is a problem there

Also I'd recommend taking a look at the crosstrack plots to make sure we haven't somehow reversed polarity of that controller

aidnem:
@jkleiber it was defaulting to algae because of no scoring. It works reliably (about an inch of error) so I'm tuning it and lineup)

jkleiber:
Nice, that's great to hear

jkleiber:
@aidnem did y'all tune these gains with the weight installed in the robot?

aidnem:
Yes but the weight is 10 lbs too light and the gains are pretty bad (mostly heading controller is bad).

jkleiber:
We should consider using the same heading control we use for lineup, that tends to work decently well

aidnem:
@jkleiber that seems like a good idea. 

One issue we had was it would overshoot its pose (sometimes), then back up, then lineup. Probably a lower acceleration or top speed would fix this (or better linear PID).

We also discovered that left camera is twisted in its mount so lineup is a little off on the left camera right now.

jkleiber:
@aidnem when going to the reef, we probably should switch to lineup much earlier than we currently are (or else we will get an overshoot)

If we're using the 2 phase system, we're guaranteed to be driving along our goal heading so we could probably make that tolerance fairly high, maybe back up to the 0.5 meters

The camera being twisted is going to have an effect, so good to find that now

src/main/java/frc/robot/subsystems/drive/states/LinearDriveState.java
aidnem: @minhnguyenbhs I fixed this yesterday afternoon, all the PIDs get applied in onEntry. This might cause a bit of lag, I think at comp we probably just use the java constants so we don't have to recreate the controllers in onEntry.

src/main/java/frc/robot/subsystems/drive/states/LinearDriveState.java
```diff
@@ -0,0 +1,328 @@
...
+    // HEADING
+    double headingError = currentPose.getRotation().minus(goalPose.getRotation()).getRadians();
+    double headingVelocity =
+        headingController.calculate(
+            currentPose.getRotation().getRadians(), goalPose.getRotation().getRadians());
+
+    double driveVelocityScalar =
+        driveController.calculate(
+            distanceToGoal,
+            new State(0.0, JsonConstants.drivetrainConstants.kDriveToPointEndVelocity));
```
jkleiber, line 300: @aidnem the most straightforward way to try to find the trapezoid profile rather than using the PID feedback would be to get the setpoint's velocity from the profiled PID: https://docs.wpilib.org/en/stable/docs/software/advanced-controls/controllers/profiled-pidcontroller.html#getting-using-the-setpoint

But your implementation in #216 is probably more accurate since you are getting the robot's actual speed and using that initial condition

aidnem: I see. The intent with this line was just to have an end velocity specified even though we were still using PID. I'll try the setpoint thing and see if it's better than my velocity calculations, since my velocity math didn't seem to work properly.

jkleiber: Yeah this is fine as it is. What I'm trying to say is that after running calculate() we can pull the velocity rather than using the output of calculate (which is the PID version) if we want to

aidnem: Yep. I'll try that too, it was really buggy on my branch but maybe it'll be better here, I might've been doing it wrong. The velocity setpoint was always really low.

aidnem:
@jkleiber I think I just realized why it was overshooting and driving back: Since we are passing in a distance which is decreasing, our velocity is actually negative when moving toward the goal. I was giving a positive ending velocity to the profiled controller, so it was probably trying to end moving the opposite direction (which would require hitting the goal the first time with negative velocity while you accelerate so that you can go back and end with positive velocity at the goal). Today at shop we should try it with a negative end velocity and see if it still does the loop-de-loop

jkleiber:
@aidnem have we plotted the velocity to see that it is commanding negative velocity. If it is, then I think you're right

aidnem:
I didn't plot it (I forgot bc I'm a little silly) but from watching it next to odometry view and in sim it's been negative. Here's a screenshot of the velocity as it drives to position in sim:

image

Distance is yellow, velocity is blue. You can see velocity is negative for the duration of the path.

aidnem:
@jkleiber to mitigate the immediate rotation causing us to bump the coral station wall on our way to the 4th score, do we want to just enable reef orbit while in phase 1 and then switch to rotation controller upon entering phase 2? My concern is that this might have the same problem.

Option 2 would be disabling rotation control when more than a certain distance away from the target (we just take the distance from coral station pose to reef 6 pose and subtract like half a meter and call it a day).

jkleiber:
@aidnem if we do reef orbit mode in phase 1 I think it'll work because we will slowly transition toward facing the correct direction and then snap to the reef face in phase 2. I've seen other teams do this successfully so let's try it

src/main/java/frc/robot/subsystems/drive/states/LinearDriveState.java
```diff
@@ -0,0 +1,347 @@
...
+
+    // Determine which pose to aim at. However, always use the distance to the final pose to compute
+    // the target speed.
+    Pose2d currentGoalPose = phase1Pose;
+    // Use whichever distance we're going to so that we slow down on the way to phase 1 pose
+    double distanceToCurrentGoal = distanceToPhase1Pose;
+    double endVelocityGoal = 0.0; // Try to reach 0 velocity in phase 1
+    if (hasEnteredPhase2
+        || distanceToGoal < JsonConstants.drivetrainConstants.kDriveToPointPhase2Distance
+        || Math.abs(angleToTarget) < JsonConstants.drivetrainConstants.kDriveToPointPhase2Angle) {
```
avidraccoon, line 280: I dont think angle to target can ever be negative

src/main/java/frc/robot/subsystems/drive/states/LinearDriveState.java
```diff
@@ -0,0 +1,347 @@
...
+  }
+
+  private double driveVelocityScalar = 0.0;
+
+  @Override
+  public void periodic() {
+    Pose2d currentPose = drive.getPose();
+
+    // Project the goal pose backwards to find the phase 1 goal pose.
+    double phase1OffsetX =
```
avidraccoon, line 250: Is there a reason this is recalculated every periodic?

src/main/java/frc/robot/subsystems/drive/states/LinearDriveState.java
jkleiber: You are correct, we could remove the abs()

src/main/java/frc/robot/subsystems/drive/states/LinearDriveState.java
jkleiber: Mainly a habit of mine because I work in real time systems where values could update every loop. But for this we could move it to onEntry. I recommend no change though since the computational overhead is minimal

aidnem:
After this update it looks like it still hits the coral station with its corner after intaking, the reef center is at too high of an angle to the OTF pose for reef 6 I think. We can test further in shop.

jkleiber:
@aidnem yeah that makes sense. I do want to avoid sudden turning motion near the reef when the elevator is extended though, so maybe this ends up being acceptable. If not, we could add a rate limit to the heading command

PR #216, opened 2025-03-27
Trapezoidal linear drive

aidnem:
Same as the other 2 linear drive PRs except it uses a trapezoid profile for the drive velocity scalar.

This seems to work but I don't trust it because it drives at snail pace and doesn't slow down near the target.


aidnem:
@jkleiber I just fixed this in case we want to go with this one instead, it might be less jank.

https://github.com/user-attachments/assets/134a86d1-783d-4545-9efe-36c284bf4709

Lmk if you need it in H264, my laptop is crashing out over the video but maybe yours can play it

src/main/java/frc/robot/subsystems/drive/states/LinearDriveState.java
```diff
@@ -0,0 +1,373 @@
...
+        VecBuilder.fill(chassisSpeeds.vxMetersPerSecond, chassisSpeeds.vyMetersPerSecond);
+
+    double velocityToGoal =
+        findVelocityTowardPoint(
+            velocity, currentPose.getTranslation(), currentGoalPose.getTranslation());
+
+    double driveVelocityScalar =
+        driveProfile.calculate(
+                0.02,
+                new State(distanceToGoal, -Math.abs(velocityToGoal)),
```
jkleiber, line 341: Why is this `-Math.abs(velocityToGoal)`? Like, why is the negative sign on the outside of this (we'll always have a negative initial velocity with this)?

src/main/java/frc/robot/subsystems/drive/states/LinearDriveState.java
```diff
@@ -0,0 +1,373 @@
...
+
+    // Calculate theta, the angle between the speakerAngle and the current velocity angle
+    Rotation2d theta = velocityAngle.minus(targetAngle);
+
+    // Project the velocity vector onto the speaker vector with ||u|| * cos(theta)
+    // Where u is the velocity vector
+    double velocityToGoal = velocity.norm() * theta.getCos();
+
+    return velocityToGoal;
+  }
```
jkleiber, line 276: comments should be updated to 2025 game lol

jkleiber (commented):
the video looks good @aidnem - it might be worth testing this on the robot tonight just to see if it works better. I left a couple comments

aidnem:
@jkleiber I forgot to push my changes :(. The video is from assuming zero velocity initially and then blindly trusting the profile. I've pushed those changes now.

jkleiber:
Closing in favor of #215

PR #217, opened 2025-03-28
Tune climb better

aidnem:
Unfortunately @j0shuaS lost the use of his thumb due to a tragic fall, so I had to take this over :(

- Moves climb FF logic to each individual state
- Adds a feedforward that boosts searching state voltage when below a certain margin away from deployed angle


aidnem:
@PChild I'm not sure if anybody from controls is at shop, but you can test this pretty easily. Literally just pull the code, deploy it, and then press climb button and it should "just work" (of course be careful because if I did something really dumb it might bend climber mount again).

src/main/deploy/constants/comp/ClimbConstants.json
```diff
@@ -1,22 +1,29 @@
 {
   "restingAngle": {
-    "value": 0.3,
+    "value": 0.0,
```
jkleiber, line 3: why would we change this value?

aidnem: Oops I wrote out constants and changed those by accident, I'll fix really quick

src/main/deploy/constants/comp/ClimbConstants.json
```diff
@@ -1,22 +1,29 @@
...
     "unit": "Radian"
   },
   "searchingAngle": {
     "value": 1.8,
     "unit": "Radian"
   },
   "finalHangingAngle": {
-    "value": 0.7,
+    "value": 0.1,
```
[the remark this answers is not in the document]
aidnem: @jkleiber thanks so much for catching this, I looked at the diff after automatically writing constants but somehow didn't see all of those changes.

jkleiber:
This is ready to test 👍

aidnem:
@jkleiber I won't be available anymore, so if you happen to find any terrible logic issues, feel free to fix them. I won't be responsive for a while.

jkleiber:
@aidnem in states where feedforward isn't used we need to explicitly set it to 0. Otherwise it could remain stale and fight the PID control I think?

jkleiber (approved):
Approved

Note: we now bypass the waiting state. The driver is responsible for making sure they don't climb before the ramp flips over. This is basically how things already work

PR #219, opened 2025-03-29
Add crossbar protection again

aidnem:
- If the elevator is within the collision range for crossbar or if it will pass through that range:
 - If wrist is below the minimum angle it is safely above, clamp it below
 - Otherwise, clamp it above

This will work unless we have a setpoint that requires the elevator to be in the danger zone while the wrist is out. In this case I'll have to add logic that lets the wrist come out before the elevator goes into the danger zone.


aidnem:
@jkleiber this will work except for the edge case I described, I can add the logic for the other stuff but I wanted you to be able to look over this if you wanted to

aidnem:
Ok wait this logic is actually horrendous and awful. Here's the new plan:

Since wrist is the priority here (wrist needs to be able to go to its setpoint, elevator should accommodate it), we should do the following

- If wrist is in a position where it can collide or if it will move through the zone where it can collide, hold the elevator on the same side of the collision zone as it already is.

aidnem:
New (final) logic is in, this is ready for review.

src/main/java/frc/robot/subsystems/scoring/ScoringSubsystem.java
```diff
@@ -867,13 +868,44 @@ public void determineProtectionClamps() {
...
+            && wristGoalAngle.gt(JsonConstants.wristConstants.crossbarBottomCollisionAngle);
+    boolean willPassDangerZoneDown =
+        wristAngle.gt(JsonConstants.wristConstants.crossbarTopCollisionAngle)
+            && wristGoalAngle.lt(JsonConstants.wristConstants.crossbarTopCollisionAngle);
+    boolean willPassDangerZone = willPassDangerZoneUp || willPassDangerZoneDown;
+
+    boolean canWristHitCrossbar = wristInDangerZone || willPassDangerZone;
+
+    if (canWristHitCrossbar) {
+      // If the wrist can hit the crossbar, keep elevator on the same side of the crossbar that it
```
jkleiber, line 885: What happens if we the wrist is in the danger zone and needs to be in the danger zone on the other side of the crossbar? Is that possible/how do we handle that

aidnem: @jkleiber currently I don't think that's possible, but it would cause an issue. I suppose I need to make this a double edged sword where if the elevator is trying to pass the danger zone I hold the wrist in a safe position too.

jkleiber: We can go with this logic for now I think if we verify that there aren't any scoring or algae intake setpoints in the danger zone. I think this is easy to understand as a restriction for now and then if needed we can add extra logic on top of this

jkleiber:
@aidnem the only notable code problem from the end of the night is that for some reason we were moving on from scoring to intake in auto way too early

The L4 setpoint needs a bit of work, and the collision avoidance logic results in us not dropping the elevator that fast, but that makes sense based on the code implementation

aidnem:
@jkleiber by only letting the wrist flip from "up" to "down" when the elevator is between the min height where it won't crush claw under it and the max height where it doesn't risk destroying itself on carriage, and forcing the elevator down into this range whenever the wrist "wants" to flip sides, we can make algae work with the current geometry.

jkleiber:
@aidnem I think they are just going to cut the claw down and we may not even need crossbar avoidance. If that's the case we'll set the cross bar height range very underground and call it good

aidnem:
Oh just like leave it where it is and chop off part of it?

jkleiber:
Yeah it should save us like 3 inches

jkleiber (approved):

PR #220, opened 2025-03-30
Add wrapper command to allow otf pathfinding to start and not start driving

avidraccoon:


jkleiber:
just fyi: before reviewing this we need some verification evidence

PR #224, opened 2025-04-01
Seed start position in auto

minhnguyenbhs:
What kind of screenshot would you like as evidence? I think the estimates for seeded setpoints might be off but the framework is hopefully done!

### DIFF
--- a/src/main/deploy/constants/comp/FeatureFlags.json
+++ b/src/main/deploy/constants/comp/FeatureFlags.json
@@ -1,7 +1,7 @@
 {
   "runDrive": true,
-  "runVision": true,
-  "runScoring": false,
+  "runVision": false,
+  "runScoring": true,
   "runClimb": true,
   "runRamp": true
 }
--- a/src/main/java/frc/robot/StrategyManager.java
+++ b/src/main/java/frc/robot/StrategyManager.java
@@ -1,5 +1,7 @@
 package frc.robot;
 
+import edu.wpi.first.math.geometry.Pose2d;
+import edu.wpi.first.math.geometry.Rotation2d;
 import edu.wpi.first.networktables.BooleanPublisher;
 import edu.wpi.first.networktables.DoublePublisher;
 import edu.wpi.first.networktables.DoubleSubscriber;
@@ -8,6 +10,7 @@ import edu.wpi.first.networktables.NetworkTableInstance;
 import edu.wpi.first.networktables.StringPublisher;
 import edu.wpi.first.networktables.StringSubscriber;
 import edu.wpi.first.wpilibj.DriverStation;
+import edu.wpi.first.wpilibj.DriverStation.Alliance;
 import edu.wpi.first.wpilibj2.command.Command;
 import edu.wpi.first.wpilibj2.command.CommandScheduler;
 import frc.robot.commands.strategies.AutoIntake;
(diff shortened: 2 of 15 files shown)


linglejack06:
> What kind of screenshot would you like as evidence? I think the estimates for seeded setpoints might be off but the framework is hopefully done!

@minhnguyenbhs maybe featureflag off vision (if thats possible?) and then run an auto without it in sim and video the proof

src/main/java/frc/robot/StrategyManager.java
```diff
@@ -435,6 +440,21 @@ public void autonomousInit(AutoStrategy strategy) {
...
 
+    if (!FeatureFlags.synced.getObject().runVision
+        || drive.getVisionMeasurementCount() < VisionConstants.synced.getObject().safetyThreshold) {
+      Pose2d startPose =
+          new Pose2d(strategy.startSeededPositionTranslation, strategy.startSeededPositionRotation);
+      if (DriverStation.getAlliance().isPresent()
+          && DriverStation.getAlliance().get() == Alliance.Blue) {
+        startPose =
+            new Pose2d(
+                17.5 - startPose.getX(),
```
linglejack06, line 451: i think the 17.5 should be 17.3736 for full accuracy? same with the 8 (should be 7.9248)

linglejack06 (changes requested):
this looks great, i can approve as long as we have some sort of visual proof or we test in shop. But code implementation looks good to me other than the numbers regarding flipping pose (there might actually be a flip pose method too? not sure)

src/main/java/frc/robot/StrategyManager.java
minhnguyenbhs: ahh i don't super know how to provide visual proof :( this was with the camera off? green is where the robot thinks it is (I can fix maplesim too if that's a high priority)

image

avidraccoon: MapleSim vision is bad due to a stupid mistake I made

avidraccoon: maplesim vision uses odometry position for one camera and maplesim position for the other camera

minhnguyenbhs: oh this doesn't use camera though so that's ok :D although I can fix that as part of this branch if you want

src/main/deploy/constants/comp/FeatureFlags.json
```diff
@@ -1,5 +1,8 @@
 {
   "runDrive": true,
+  "runVision": false,
```
jkleiber, line 3: note: we need to turn this back on before merging

src/main/java/frc/robot/constants/VisionConstants.java
```diff
@@ -24,4 +24,6 @@ public class VisionConstants {
           Units.inchesToMeters(-5.5),
           Units.inchesToMeters(12.0),
           new Rotation3d(0, 0, 0));
+
+  public final Integer safetyThreshold = 10;
```
jkleiber, line 28: Could we rename this something like "minimumVisionMeasurementInitCount" or similar to indicate that this is a counter of how many vision measurements we need to trust that the robot is in the correct location?

src/main/java/frc/robot/StrategyManager.java
```diff
@@ -435,6 +440,21 @@ public void autonomousInit(AutoStrategy strategy) {
...
+        || drive.getVisionMeasurementCount() < VisionConstants.synced.getObject().safetyThreshold) {
+      Pose2d startPose =
+          new Pose2d(strategy.startSeededPositionTranslation, strategy.startSeededPositionRotation);
+      if (DriverStation.getAlliance().isPresent()
+          && DriverStation.getAlliance().get() == Alliance.Blue) {
+        startPose =
+            new Pose2d(
+                17.3736 - startPose.getX(),
+                7.9248 - startPose.getY(),
+                new Rotation2d(startPose.getRotation().getRadians() - Math.PI / 2));
```
jkleiber, line 453: can we add verification that this works when we are on the blue alliance? I think a screenshot of where the robot thinks it starts in auto would be fine for this for blue

jkleiber (commented):
first round feedback

src/main/java/frc/robot/StrategyManager.java
minhnguyenbhs: image

Issue #225, opened 2025-04-01
Fix phase 1 to 2 overshoot in Linear Drive


aidnem:
**Purpose**
Currently linear drive ignores the fact that it has to turn between phases 1 and 2, so it blasts through the change at full speed ans completely misses. It needs to slow down before the end of phase 1.

**Project Scope**
- [ ] Figure out how to do this (maybe look at old mentor built branch)

aidnem:
I've fixed this on the mentor run branch; I'm going to branch off of that, interactive rebase off main, and then hopefully we can merge in this fix

Here's verification:

![Image](https://github.com/user-attachments/assets/e4f2dfa4-f7d9-4826-b453-471d955e9f6e)

Yellow is setpoint, blue is actual; you can see that yellow is always under blue so we are always moving in the correct direction.

Here's drive velocity scalar, you can see it is always negative:

![Image](https://github.com/user-attachments/assets/a00d73aa-ebeb-4ec3-9f69-0b9018e47550)

jkleiber:
@aidnem thanks for alleviating my concerns - this looks smooth. We should test this once we get the elevator setpoints locked in (I think we are probably okay on L2 and L3 for now, just need to get L4 and we can get the algae targets if needed when we do those PRs)

jkleiber:
The L4 setpoint we had last night was working decently with the coral fully pointing away from us. I think we may need to lift the elevator a bit higher but it should work well

aidnem:
Nice

aidnem:
One positive is that it looks like the overshoot was actually fixed by this branch

aidnem:
@jkleiber thanks for letting me know. How are autos/our ability to score looking?

jkleiber:
We can score and at the end of the night we probably would have had a 3 piece auto if the intake station was wider

PR #226, opened 2025-04-01
Fix linear drive overshoot

aidnem:
Begin by profiling linear drive with phase 1 distance in phase 1, then switch to phase 2 profile using bumplessing for a (hopefully) smooth motion.


aidnem:
@jkleiber personally I believe this to be a priority to test in shop as soon as we merge the first scoring PR, I think avoiding flipping the robot over is going to be pretty vital. I'll try to definitively prove that it works in sim (I was able to reproduce the driving backward issue in sim, so I trust fixing it) so that we don't have to waste time debugging it on the robot. I've already provided verification on #225 but I'll try to test more cases to make sure I haven't missed anything.

src/main/java/frc/robot/subsystems/drive/states/LinearDriveState.java
```diff
@@ -270,11 +287,41 @@ public void periodic() {
...
+          double lastVelocity = lastState.velocity;
+          double lastError = distanceToCurrentGoal - lastState.position;
+          double adjustedPosition = distanceToGoal - lastError;
+          driveController.reset(new State(adjustedPosition, lastVelocity));
+
+          System.out.println("Switched to phase 1 with bumplessing");
+        } else {
+          driveController.reset(new State(distanceToGoal, 0.0));
+
+          System.out.println("Switched to phase 1 instantly");
```
jkleiber, line 315: These should say "switched to phase 2" right?

jkleiber (commented):
One nitpick comment but looks good to test

aidnem:
@jkleiber it won't let me approve because I'm the PR author, but I looked over all the changes you made and they make sense (looks like just a copy paste of init state).

jkleiber (approved):

Issue #229, opened 2025-04-02
Lower climb feedforward 


jkleiber:
Now that the climb motors aren't doing weird things, we should reduce the feedforward in both directions because it's too fast

Lets start with 8V and see how that goes

PR #231, opened 2025-04-02
(#229) Decrease climb feedforwards

aidnem:
Climb was moving way too fast after we fixed the climb jitter issues. Decreasing the feedforwards should stop it from unwinding or snapping the rope.


jkleiber (commented):
I think this looks good

jkleiber (approved):

PR #232, opened 2025-04-03
Add barge algae auto

aidnem:
Adds a flag to AutoStrategy that, when true, ignores the strategy you have and replaces it with:

- Score 1 coral on reef 0 L4
- Drive slowly back to auto line
- Intake L2 algae while driving slowly forward from auto line
- Drive slowly back to auto line
- Score net


src/main/java/frc/robot/constants/subsystems/DrivetrainConstants.java
```diff
@@ -252,6 +252,9 @@ public class DrivetrainConstants {
   public final Double kPositionTolerance = 0.25;
   public final Double kVelocityTolerance = 0.1;
 
+  public final Double kDriveTranslationMaxVelocitySlow = 1.5;
+  public final Double kDriveTranslationMaxAccelerationSlow = 1.0;
```
jkleiber, line 256: This seems too slow imo, we could probably drive at like 2 m/s and do 3 m/s^2. We can build up on the practice field though

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -705,6 +707,17 @@ public void setDesiredLocation(DesiredLocation location) {
...
       this.desiredLocation = location;
     }
+
+    if (location == DesiredLocation.AutoLine) {
+      System.out.println("IT GOT SET");
+      System.out.println("IT GOT SET");
+      System.out.println("IT GOT SET");
+      System.out.println("IT GOT SET");
+      System.out.println("IT GOT SET");
+      System.out.println("IT GOT SET");
```
jkleiber, line 717: Can we make this only have <=1 println

src/main/deploy/auto/1PieceBargeWithAlgae.json
```diff
@@ -0,0 +1,11 @@
+{
+  "scoringLocations": [
+    "Reef0"
+  ],
+  "scoringLevels": [
+    "L4"
+  ],
```
jkleiber, line 7: note: these currently do nothing

src/main/java/frc/robot/StrategyManager.java
```diff
@@ -185,6 +188,26 @@ public void logActions() {
...
+          new Action(
+              ActionType.IntakeAlgae, GamePiece.Algae, DesiredLocation.Algae0, FieldTarget.L2));
+      this.addAction(
+          new Action(
+              ActionType.DriveToLine, GamePiece.Algae, DesiredLocation.AutoLine, FieldTarget.L2));
+      this.addAction(
+          new Action(
+              ActionType.NetScore, GamePiece.Algae, DesiredLocation.AutoLine, FieldTarget.Net));
+
+      return;
```
jkleiber, line 208: we should add another drive to algae action to ensure we get move points

jkleiber (commented):
first pass comments

aidnem:
@jkleiber I think I'll complete this and then we can replace it with correct StrategyManager implementation later. The new desired locations will provide a decent starting point so we don't have to write that stuff again.

aidnem:
Something is incredibly suspicious here:
image

It looks like AutoIntakeBargeAlgae's end method:

image

is getting called AFTER the init of AutoDriveToLocation and therefore it's cancelling the linear drive:

image

aidnem:
In an insane turn of events, the command scheduler was only ending the previous command after we scheduled the next one. Fixed by just adding a manual call to cancel in strategy manager so that it cancels commands whenever they report that they are finished so that they don't linger until the next CommandScheduler loop and destroy robot state with their end() methods.

jkleiber:
Nice find @aidnem !

aidnem:
Still no luck on making the final drive to position command work, its ending way too early. I'm very close though, I added a ton of logging and I'm 90% sure it's ending before it can properly initialize so its goal pose is really cooked. I can try more over the weekend.

aidnem:
Apparently you can break wiring in sim too?? not sure what's happening but I'm going to restart my laptop and maybe give up for today

image

aidnem:
https://github.com/user-attachments/assets/7d7a1a80-f194-45a0-b5c0-4e7f57e78462

This video is from before I fixed the elevator staying up bug but it shows everything else working flawlessly :)

src/main/java/frc/robot/subsystems/drive/Drive.java
```diff
@@ -717,6 +727,10 @@ public void setDesiredLocation(DesiredLocation location) {
     if (isLocationScoring(location)) {
       this.desiredLocation = location;
     }
+
+    if (location == DesiredLocation.AutoLine) {
+      this.desiredLocation = location;
+    }
```
jkleiber, line 733: Can we make this an else if?

src/main/java/frc/robot/subsystems/scoring/ClawMechanism.java
```diff
@@ -23,13 +23,11 @@ public class ClawMechanism {
 
   private Debouncer algaeRiseDebouncer =
       new Debouncer(
-          JsonConstants.clawConstants.algaeCurrentDetectionTimeRising.in(Seconds),
-          DebounceType.kBoth);
+          JsonConstants.clawConstants.algaeDetectionTimeRising.in(Seconds), DebounceType.kBoth);
```
jkleiber, line 26: This and the other debouncer are both kBoth here, which seems sus

jkleiber (commented):
just some code comments

src/main/java/frc/robot/subsystems/scoring/ClawMechanism.java
aidnem: I think our original thought was that, since rise time was faster than falling time, using an OR between the two debouncers would give us an asymmetric debouncer without having to use rising or falling mode. I'll fix that now.

jkleiber:
@aidnem verification evidence from shop

jkleiber (approved):

PR #238, opened 2025-04-16
pick an algae off the reef in auto- logic

minhnguyenbhs:
I haven't been able to run the full auto sequence yet because of lineup not working. The testing auto is called RW- thanks!!


jkleiber (commented):
I think this looks great @minhnguyenbhs - we will need to update all the existing autos though to use the new schema, so let's do that in this PR as well

linglejack06 (approved):
lgtm

jkleiber (approved):

PR #251, opened 2025-08-22
Add README

aidnem:
Adds README.md with instructions on how to use the sim, a description of impressive features of this year's codebase, and a general description the project's structure.


aidnem:
@godmar I've written instructions on how to get started with the sim. The rest of the README isn't done yet, so this PR is a draft until I finish that, but I wanted you to have access to the guide as soon as possible. Please let me know about anything I can improve in the guide, I want it to be as helpful as possible.

aidnem:
You're welcome!
Before each match we would take a whiteboard with a top-down view of the field and draw out each of our autos to make sure we didn't collide. We would also generally plan who would score where in teleop.

PR #254, opened 2025-10-13
Separate lineup offsets in constants

### LINKED ISSUE #252: Add lineup heading offset and individual along track offsets per camera
**Purpose**
Our camera mounts are more bent than ever. We need not only to have an along track and crosstrack offset but also an offset for rotation, since we now have a heading error of ~5 degrees in one mount.

The mounts are now also different enough from each other that we need separate along track offsets for each camera. 

**Project Scope**
- [ ] Separate Lineup Along Track offset into 2 variables, one for left and 1 for right
- [ ] Add a rotation offset for left and a rotation offset for right
- [ ] Verify this by doing some autonomous scoring (Smart mode)

aidnem:
Allow separate AlongTrack offsets per camera and add a heading offset for each camera when scoring in lineup.


src/main/java/frc/robot/subsystems/drive/states/LineupState.java
```diff
@@ -460,9 +479,20 @@ public void LineupWithReefLocation() {
               * getAlongTrackVelocityReductionFactor(observation.crossTrackDistance())
               * getAlongTrackVelocity(observation.alongTrackDistance()); // Maybe use filtered?
       double vy = driveCrossTrackLineupController.calculate(observation.crossTrackDistance());
+
+      Angle headingOffset;
+      if (cameraIndex == 0) {
```
godmar: Same here.

src/main/deploy/constants/comp/DrivetrainConstants.json
godmar: These changes could break the `demo`, `hitl`, and `test_drivebase` configurations which rn have `driveAlongTrackOffset` values.

godmar (dismissed):
I checked for other uses of `driveAlongTrackOffset` (found some in `demo`, `hitl`, `test_drivebase`), and discovered the `*CameraIndex` constants which might be applicable here. Decide if this is worth addressing now.

src/main/java/frc/robot/subsystems/drive/states/LineupState.java
```diff
@@ -337,13 +340,23 @@ public DistanceToTag tryOtherCamera(
             ? 0
             : JsonConstants.visionConstants.FrontLeftTransform.getY()
                 - JsonConstants.visionConstants.FrontRightTransform.getY();
+
+    double alongTrackOffset;
+    if (otherCameraIndex == 0) {
```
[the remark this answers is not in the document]
aidnem: Very interesting! I suppose all of our offsets were flipped then. Glad it was only a difference of 1cm or we may have noticed because we would miss every shot 🤣 . Thanks for catching this, I've updated it now.

aidnem:
Thank you for the quick, short-notice review, by the way!

godmar (approved):
LGTM

